PR #27, opened 2020-03-26
Fix root dir of function config

wolfstudy:

- Fix root dir of function config


charts/pulsar/templates/broker-configmap.yaml
```diff
@@ -55,7 +55,7 @@ data:
   PF_containerFactory: k8s
   PF_numFunctionPackageReplicas: "{{ .Values.broker.configData.managedLedgerDefaultEnsembleSize }}"
   # support version >= 2.5.0
-  PF_functionRuntimeFactoryConfigs_pulsarRootDir: /pulsar
+  PF_functionRuntimeFactoryConfigs_pulsarRootDir: /sn-platform
```
sijie: @wolfstudy we should support both sn-platform and open-source pulsar.

PR #35, opened 2020-04-02
Support logger level setting

wolfstudy:

Fixes: #https://github.com/apache/pulsar/issues/5917

- Support logger level setting

The `apply-log-from-env.py` at https://github.com/apache/pulsar/pull/6652.


wolfstudy:
> 1. let's separate the change of bumping version from adding configuration.
> 2. for configuring the logging level, we should use `PULSAR_LOG_*` environment variables.
> 3. users should be able to configure the logging level for individual components.

Thanks @sijie , Currently, we use mount `configmap` way, if user want to config log level, they can modify the `log4j-configmap.yaml` file.

PTAL again thanks.

charts/pulsar/templates/broker-statefulset.yaml
```diff
@@ -169,6 +169,7 @@ spec:
           echo "OK" > status;
           {{- include "pulsar.broker.zookeeper.tls.settings" . | nindent 10 }}
           bin/pulsar zookeeper-shell -server {{ template "pulsar.zookeeper.connect" . }} get {{ template "pulsar.broker.znode" . }};
+          cp /logconf/log4j2.yaml /sn-platform/conf;
```
sijie, line 172: why not just mount the log4j2.yaml directly to the target log4j2 configuration file without copying the file?

sijie, line 172:
```suggestion
          cp /logconf/log4j2.yaml {{ template "pulsar.home" . }}/conf;
```

Please use a template if possible.

charts/pulsar/templates/log4j-configmap.yaml
```diff
@@ -0,0 +1,143 @@
...
+metadata:
+  name: "{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}-log4j"
+  namespace: {{ .Values.namespace }}
+  labels:
+    {{- include "pulsar.standardLabels" . | nindent 4 }}
+    component: {{ .Values.broker.component }}
+data:
+  # Include prometheus configuration file, setup to monitor all the
+  # Kubernetes pods with the "scrape=true" annotation.
+  log4j2.yaml: |
```
sijie, line 31: I don't think we need to add a separate config map for each component. We should just add the `log4j2.yaml` field to the existing configmap for each component.

charts/pulsar/templates/broker-statefulset.yaml
wolfstudy: If I directly mount the `log4j2.yaml` file under `/sn-platform/conf`, the other files in this folder will be overwritten.

wolfstudy: Good idea

charts/pulsar/templates/log4j-configmap.yaml
wolfstudy: I don't think this will have any impact. From the user's point of view, I think it will be more friendly. Do we provide a unified configuration place for the log, or do we disassemble it to configure it. I prefer the former.

wolfstudy: I don't think this will have any impact. From the user's point of view, I think it will be more friendly. Do we provide a unified configuration place for the log, or do we disassemble it to configure it. I prefer the former.

wolfstudy:
@sijie PTAL again thanks

charts/pulsar/templates/bookkeeper-statefulset.yaml
```diff
@@ -133,6 +133,7 @@ spec:
         - >
           bin/apply-config-from-env.py conf/bookkeeper.conf;
           bin/apply-config-from-env.py conf/pulsar_env.sh;
+          cp /logconf/log4j2.yaml {{ template "pulsar.home" .}}/conf;
```
sijie, line 136: Why not just mount the config map directly to the target path?

charts/pulsar/templates/broker-statefulset.yaml
```diff
@@ -194,9 +195,14 @@ spec:
...
             name: "{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}"
         volumeMounts:
-          {{- include "pulsar.broker.token.volumeMounts" . | nindent 10 }}
-          {{- include "pulsar.broker.certs.volumeMounts" . | nindent 10 }}
+          - mountPath: "/logconf"
+            name: "{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}-log4j-config"
+        {{- include "pulsar.broker.token.volumeMounts" . | nindent 10 }}
+        {{- include "pulsar.broker.certs.volumeMounts" . | nindent 10 }}
       volumes:
+      - name: "{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}-log4j-config"
```
sijie, line 203: volume and volumeMount are required for different components. why not just create templates in `__helper.tpl`? Just like what we did for `pulsar.broker.token.volumes`?

charts/pulsar/templates/log4j-configmap.yaml
```diff
@@ -0,0 +1,143 @@
+#
```
sijie, line 1: It seems that this log4j-configmap file is used by multiple components. If we change the log setting in this log4j configmap, it will affect all the components. This doesn't make sense to me. We should be able to configure log4j for individual components.

charts/pulsar/templates/bookkeeper-statefulset.yaml
wolfstudy: Because the `/sn-platform/conf` folder already exists, if you mount the path here, all the original files will be overwritten.

charts/pulsar/templates/log4j-configmap.yaml
wolfstudy: Thanks @sijie will fix it

charts/pulsar/templates/broker-statefulset.yaml
wolfstudy: Good idea, will fix it

wolfstudy:
@sijie done, PTAL again thanks

sijie:
@wolfstudy I changed a bit the way how do you mount the log config yaml file and add the ability to support automatically rolling deployments when the configmap is changed.

Issue #33, opened 2020-03-29
Presto Deployment on K8s


eaba:
To deploy presto, I have this:
````
bin/apply-config-from-env.py conf/presto/config.properties
                    bin/apply-config-from-env.py conf/presto/catalog/pulsar.properties
                    bin/pulsar sql-worker run
````
I get this error:
````
File "bin/apply-config-from-env.py", line 50, in <module>
    k,v = line.split('=', 1)
ValueError: need more than 1 value to unpack
````

eaba:
I was able to get some good result here: https://github.com/eaba/presto-pulsar-chart

PR #37, opened 2020-04-04
Support presto

eaba:


charts/pulsar/values.yaml
```diff
@@ -83,13 +83,15 @@ components:
...
+  # pulsar sql
+  presto: true
 
 ## Monitoring Components
 ##
 ## Control what components of the monitoring stack to deploy for the cluster
 monitoring:
   # monitoring - prometheus
-  prometheus: true
+  prometheus: false
```
sijie, line 94: Can you leave the setting unchanged?

charts/pulsar/values.yaml
```diff
@@ -609,14 +615,14 @@ broker:
   # use a component name that matches your grafana configuration
   # so the metrics are correctly rendered in grafana dashboard
   component: broker
-  replicaCount: 3
+  replicaCount: 2
```
sijie, line 618: keep this setting as default?

eaba: @sijie , does the new commits satisfy your request?

.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
sijie, line 1: Can you exclude `.vs` from the pull request?

charts/pulsar/values.yaml
```diff
@@ -113,6 +115,10 @@ images:
     repository: streamnative/platform
     tag: v1.0.0
     pullPolicy: IfNotPresent
+  presto:
+    repository: eabanonu/presto-pulsar
```
sijie, line 119: any reason why you can't use `apachepulsar/pulsar-all`?

eaba: I have tried, there is not documentation on how to deploy pulsar sql this way - the reason for this chart.
#33 is the issue I have encountered. 
What has been done in eabanonu/presto-pulsar is to extract pulsar connector into presto plugin folder.

.vs/ProjectSettings.json
eaba: Ok

charts/pulsar/values.yaml
eaba: Let me think on this, it might be possible to use pulsar-all given the way it is currently been done

sijie: thank you @eaba!

sijie: Okay. I am fine with a customized image for now. We will look into how to improve this using apachepulsar/pulsar-all image. Thank you @eaba!

sijie: Can you remove `.vs` directory? Then I can merge this.

sijie:
@eaba awesome contributions!

PR #45, opened 2020-04-10
Rebased

eaba:


eaba:
@sijie I don't know why this is failing to install - just did another helm install at my end and all is green!

sijie:
@eaba I have merged #46 which includes your initial changes here. Thank you so much for your contributions!

PR #47, opened 2020-04-13
Update README.md

wolfstudy:


wolfstudy:
the scripts directory exists in `charts` instead of `charts/charts`

README.md
```diff
@@ -119,9 +119,6 @@ We provide some instructions to guide you through the preparation for the follow
     ```bash
     git clone https://github.com/streamnative/charts.git
     ```
-    ```bash
```
sijie, line 122: after cloning the repo, it will create a directory `charts`. So you need to enter the chart dir first, no?

wolfstudy: No, because the `scripts` directory does not exist in `charts/charts`.

wolfstudy:
yes, this right

PR #51, opened 2020-04-14
Update secret name to align with auto generate secret name

zymap:
---

*Motivation*

Make the secret name aligned with the secret names generated by the
`prepare_helm_release.sh`.


sijie:
@zymap this is about the certs. did we fix the token secrets as well?

PR #57, opened 2020-04-22
Add loki to helm

wolfstudy:

- Add loki to helm


wolfstudy:
Maybe we can replace this with helm dependency, so close this pull request.

Issue #56, opened 2020-04-20
Add Grafana Loki to the charts


codelipenghui:
There are two components in the Loki, Loki server and log agent named Promtail. All of these components support Helm installation. Here are tutorials for Loki and Promtail, https://github.com/grafana/loki/blob/v1.4.1/docs/installation/helm.md#installing-loki-with-helm. https://github.com/grafana/loki/blob/v1.4.1/docs/clients/promtail/installation.md#installing-promtail.

Loki also have a docker log driver. https://github.com/grafana/loki/tree/master/cmd/docker-driver#overview. This one is easier to use and looks don't need to install Promtail.

jiazhai:
related to migration

wolfstudy:
https://github.com/streamnative/charts/pull/57

wolfstudy:
https://github.com/streamnative/apache-pulsar-grafana-dashboard/pull/29

PR #58, opened 2020-04-23
Add install loki docs

wolfstudy:

- Add install loki docs


jiazhai:
#56

sijie:
closed this pr since loki is already added as a dependency in #63

PR #63, opened 2020-04-25
Install loki as a dependency

sijie:


sijie:
@wolfstudy I have added loki as a dependency. so we don't need #58 anymore.

PR #64, opened 2020-04-26
Fix alert manager

sijie:
*Modifications*

- upgrade prometheus to v2
- alert manager should mount config from configmap
- support config reload


charts/pulsar/templates/alertmanager-statefulset.yaml
```diff
@@ -57,27 +56,69 @@ spec:
      {{- end }}
       terminationGracePeriodSeconds: {{ .Values.alert_manager.gracePeriod }}
       containers:
+      {{- if .Values.configmapReload.alertmanager.enabled }}
```
zymap, line 59: What is this used for? How does it work?

sijie: this is used for reloading configuration without restarting the statefulset. when a configmap is updated, the `configmapReload` container is used for posting requests to reload configurations.

PR #68, opened 2020-04-27
[WIP] Add new cluster to repo

wolfstudy:

- Add new cluster to repo


wolfstudy:
This change mainly wants to migrate from an old pulsar cluster to a new pulsar cluster.

The order of migrating components:

1. wait for the zk node started by `cluster-new` to synchronize the data, and then use `reconfig` to register the zk in `cluster-new` to the zk node in `cluster-old`, so that the zk clusters of these two clusters reconstruct a new zk Cluster.
2. Uninstall broker
3. Uninstall bk

PR #85, opened 2020-05-08
Add pulsar detector to repo

wolfstudy:

- Add pulsar detector to repo

image

### DIFF
--- a/charts/pulsar/templates/prometheus-configmap.yaml
+++ b/charts/pulsar/templates/prometheus-configmap.yaml
@@ -44,7 +44,7 @@ data:
     scrape_configs:
     - job_name: 'prometheus'
       static_configs:
-      - targets: ['127.0.0.1:{{ .Values.prometheus.port }}']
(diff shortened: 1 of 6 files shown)


wolfstudy:
image

wolfstudy:
Verified in helm, everything is ok

sijie:
@wolfstudy maybe make it a different chart. it doesn't have to be deployed with the main pulsar chart.

wolfstudy:
> @wolfstudy maybe make it a different chart. it doesn't have to be deployed with the main pulsar chart.

The pulsar-detector needs to rely on the broker environment. Its main job is to detect the status of the broker. If it is deployed as a separate helm chart, how do we detect the status of the real broker we want to test?

sijie:
>  If it is deployed as a separate helm chart, how do we detect the status of the real broker we want to test?

You can specify the pulsar cluster that you want to detect. But I am fine with us starting from this approach. We can change it in the future.

PR #93, opened 2020-05-12
Add tiered storage to repo

wolfstudy:

- Add tiered storage to repo


wolfstudy:
@sijie PTAL

charts/pulsar/values.yaml
```diff
@@ -140,8 +140,8 @@ images:
   broker:
     # repository: apachepulsar/pulsar-all
     # tag: 2.5.0
-    repository: streamnative/platform
-    tag: v1.0.0
+    repository: streamnative/platform-all
+    tag: v1.0.1
```
wolfstudy, line 144: If we want to use offload, we need to use `streamnative/platform-all` image

wolfstudy:
image

image

wolfstudy:
@sijie PTAL

charts/pulsar/values.yaml
```diff
@@ -786,6 +788,23 @@ broker:
   pdb:
     usePolicy: true
     maxUnavailable: 1
+  offload:
+    enable: false
```
sijie, line 792:
```suggestion
    enabled: false
```

charts/pulsar/values.yaml
```diff
@@ -786,6 +788,23 @@ broker:
...
+    managedLedgerOffloadDriver: aws-s3
+    managedLedgerMinLedgerRolloverTimeMinutes: "1"
+    managedLedgerMaxEntriesPerLedger: "5000"
+    gcs:
+      enable: false
```
sijie, line 797:
```suggestion
      enabled: false
```

charts/pulsar/values.yaml
sijie: We used `enabled` in other places. Let's use a consistent naming convention.

charts/pulsar/values.yaml
```diff
@@ -786,6 +788,23 @@ broker:
...
+      gcsManagedLedgerOffloadRegion: pulsar-topic-offload-1
+      gcsManagedLedgerOffloadBucket: europe-west3
+      gcsManagedLedgerOffloadMaxBlockSizeInBytes: "67108864"
+      gcsManagedLedgerOffloadReadBufferSizeInBytes: "1048576"
+    s3:
+      enable: false
```
sijie, line 803:
```suggestion
      enabled: false
```

charts/pulsar/values.yaml
```diff
@@ -786,6 +788,23 @@ broker:
...
+      s3ManagedLedgerOffloadRegion: us-west-2
```
sijie, line 804:
```suggestion
      s3ManagedLedgerOffloadRegion: "[YOUR REGION]"
```

charts/pulsar/values.yaml
```diff
@@ -786,6 +788,23 @@ broker:
...
+      s3ManagedLedgerOffloadBucket: test-offload-1
```
sijie, line 805:
```suggestion
      s3ManagedLedgerOffloadBucket: "[YOUR BUCKET]"
```

charts/pulsar/values.yaml
```diff
@@ -786,6 +788,23 @@ broker:
...
     maxUnavailable: 1
+  offload:
+    enable: false
+    managedLedgerOffloadDriver: aws-s3
+    managedLedgerMinLedgerRolloverTimeMinutes: "1"
+    managedLedgerMaxEntriesPerLedger: "5000"
+    gcs:
+      enable: false
+      gcsManagedLedgerOffloadRegion: pulsar-topic-offload-1
+      gcsManagedLedgerOffloadBucket: europe-west3
```
sijie, line 799:
```suggestion
      gcsManagedLedgerOffloadBucket: "[YOUR BUCKET]"
```

charts/pulsar/values.yaml
```diff
@@ -786,6 +788,23 @@ broker:
...
     usePolicy: true
     maxUnavailable: 1
+  offload:
+    enable: false
+    managedLedgerOffloadDriver: aws-s3
+    managedLedgerMinLedgerRolloverTimeMinutes: "1"
+    managedLedgerMaxEntriesPerLedger: "5000"
+    gcs:
+      enable: false
+      gcsManagedLedgerOffloadRegion: pulsar-topic-offload-1
```
sijie, line 798:
```suggestion
      gcsManagedLedgerOffloadRegion: "[YOUR REGION]"
```

scripts/pulsar/prepare_helm_release.sh
```diff
@@ -33,6 +33,7 @@ Options:
        --control-center-password        the password of control center administrator
        --pulsar-superusers              the superusers of pulsar cluster. a comma separated list of super users.
        -c,--create-namespace            flag to create k8s namespace.
+       --service-account-key-file       the path of service account key file.
```
sijie, line 36: Isn't the service account key file only applied to GCS? Does it work for S3? If it only works for GCS, please use a more explicit name as the option name and add the document.

Besides that, since we are creating the service account key as secret for GCS, why not also provide an option to create a secret for S3?

wolfstudy: > Isn't the service account key file only applied to GCS? Does it work for S3?

Yes, the `service-account-key-file` only applied to GCS.

> Besides that, since we are creating the service account key as a secret for GCS, why not also provide an option to create a secret for S3?

In S3, we can config the secret by `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` in `values.yaml`

wolfstudy:
@sijie PTAL

wolfstudy:
ping @sijie PTAL

Issue #96, opened 2020-05-21
The grafana canot login


wolfstudy:
Version: master

Reproduce:

1. pull master code
2. kubectl create namespace pulsar
3. ./scripts/pulsar/prepare_helm_release.sh -n pulsar -k sn-platform-pulsar -c --control-center-admin admin --control-center-password adminmin
4. cd charts && helm install sn-platform-pulsar --values pulsar/values.yaml ./pulsar
5. kubectl expose service sn-platform-pulsar-grafana  --type=LoadBalancer --name=pulsar-grafana-1 --port=3000  -n pulsar
6. access grafana servie EXTERNAL-IP:3000

the error as follows:

![image](https://user-images.githubusercontent.com/20965307/82532551-340cff00-9b74-11ea-8b43-2d197449af1e.png)

PR #97, opened 2020-05-22
[Issue:96] Fix the grafana canot login

wolfstudy:

Fixes: #96 

- Fix the grafana canot login


wolfstudy:
Currently, we have changed the login method of grafana, we can configure the user and password required by grafana login in `values.yaml`

PR #101, opened 2020-06-08
Fallback to Helm .Release.Namespace when .namespace not defined

### LINKED ISSUE #100: Provide a way to use the --namespace from helm
Currently pulsar chart provides a way to specify the namespace using `.Values.namepsace`. I would propose that when the .Values.namespace is not defined, the chart falls back to using `.Release.Namespace`.

I think this would make the chart easier to reuse in other charts while still allowing the namespace to be set explicitly.

There is some discussion around best practices for helm namespaces [here](https://github.com/helm/helm/issues/5465)

enriched:
This defaults the pulsar namespace to helms default .Release.Namespace if namespace is not defined. This is implemented in the new "pulsar.namespace" template.

The .namespace is set to "pulsar" in the default values.yaml, in order to unset it `--set namespace=null` can be passed to the `helm install` command.

closes #100

### DIFF
--- a/charts/pulsar/templates/_helpers.tpl
+++ b/charts/pulsar/templates/_helpers.tpl
@@ -19,6 +19,13 @@ Expand the name of the chart.
 {{- end -}}
 
 {{/*
+Expand the name of the chart.
+*/}}
+{{- define "pulsar.namespace" -}}
+{{- default .Release.Namespace .Values.namespace -}}
+{{- end -}}
+
+{{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 If release name contains chart name it will be used as a full name.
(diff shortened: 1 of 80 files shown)


sijie:
@enriched great contribution!

austince:
Hey, thanks for this contribution! Please correct me if I'm wrong, but I think it might have missed a few instances in some of the `.tpl` files:

https://github.com/streamnative/charts/blob/58b0b8bba556fff959425b9fc229606dda77cf9d/charts/pulsar/templates/zookeeper/_zookeeper.tpl#L23-L25

I've found reference to the direct `.namespace` value in the following directory's `_{dir}.tpl`
- autorecovery
- bookkeeper
- broker
- pulsar_manager
- toolset
- zookeeper

enriched:
@austince I think you're right, thanks for catching that! If you could open a new issue that would be great.

PR #135, opened 2020-09-08
Microsoft Azure

eaba:


eaba:
Fails with this error:
```
Error: unable to build kubernetes objects from release manifest: [unable to recognize "": no matches for kind "Certificate" in version "cert-manager.io/v1", unable to recognize "": no matches for kind "Issuer" in version "cert-manager.io/v1"]
```
I don't think it is related to this pull request

sijie:
@eaba Do you want to rebase this pull request?

PR #149, opened 2020-10-23
Fix empty storageClassName issue

sijie:
*Motivation*

If no storage class name is specified, the bookkeeper statefulset will be using an empty storage class name.
It will cause pvc pending in creating pv. Because no storage class name is found.

```
storageClassName: ""
```

This change fixes the issue.


wolfstudy (approved):
LGTM +1

PR #163, opened 2020-12-11
Update kafkaPlainText label to DNS-1123 compliant label

dave-b-code:
Currently there are uppercase characters in the label which causes invalid value errors, I have converted them all to lowercase which should resolve this issue. Thanks.


sijie (approved):

sijie:
@dave-b-code thank you for catching this!

PR #173, opened 2021-01-13
Enable state storage on helm chart

sijie:


sijie:
#299 added the support. So close this issue.

PR #186, opened 2021-01-27
Allow specify different images for presto coordinator & worker

sijie:


charts/pulsar/templates/presto/presto-worker-statefulset.yaml
```diff
@@ -103,8 +103,8 @@ spec:
       {{- end }}
       containers:
         - name: {{ .Chart.Name }}-worker
-          image: "{{ .Values.images.presto.repository }}:{{ .Values.images.presto.tag }}"
-          imagePullPolicy: {{ .Values.images.presto.pullPolicy }}
+          image: "{{ .Values.images.presto_worker.repository }}:{{ .Values.images.presto_worker.tag }}"
```
addisonj, line 106: should we maybe not set a default and instead have this fall back to the existing image?

I am not sure how well that works with helm templates, but it might make it more as expected for people who don't need different images

addisonj (approved):

addisonj (approved):

PR #218, opened 2021-04-26
Fix rbac/config for namespace only prom

addisonj:
In the event of using a prometheus that can't read a whole cluster and
only a subset of data, we need to set some different rbac and config
policies.

This was started before, but now fixes a few issues with that


sijie (approved):

withernet (approved):
LGTM

PR #231, opened 2021-06-16
Mount the backup and restore configmap as environment

zymap:
---

*Motivation*

Mount backup and restore configmap as environment so user
can use configData to inject the extra environment variable.

### DIFF
--- a/.ci/release.sh
+++ b/.ci/release.sh
@@ -62,11 +62,11 @@ function release::package_chart() {
 }
 
 function release::upload_packages() {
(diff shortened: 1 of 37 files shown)


zymap:
The `Use Pulsar image` CI failed with same reason in [apache/pulsar-helm-charts](https://github.com/apache/pulsar-helm-chart/pull/130), the root cause should be this https://github.com/apache/bookkeeper/pull/2740. So I will force merge this PR

PR #232, opened 2021-06-16
Update snpe charts

rockybean:
- Fix some bug in sn-pulsar chart


rockybean:
Need to publish the latest modification to StreamNative Platform chart. So merge first and should not impact on current chart

PR #271, opened 2021-08-09
Fix the presto coordinator service name

zymap:
---

*Modifications*

Update the presto coodinator service to the corresponds service name.


zymap:
@sijie @addisonj I think we have written the wrong service name for the presto coordinator. But if I fix that by overwriting, I think it will upgrade failed from a deployed cluster. But it looks like using `presto.coordiantor` is the correct service name. What do you think? Do we need to change that?

PR #274, opened 2021-08-11
Add integration tests for the upgrade process

zymap:
**Motivation**

The streamnative/charts can upgrade from the apache/pulsar-helm-charts 
to support more components provided in the streamnative/charts.
Add an integration test for the default deploy upgrade check.


.ci/clusters/values-apache-default.yaml
```diff
@@ -0,0 +1,1019 @@
+#
```
sijie, line 1: Can you avoid adding the entire default values file? You should just add the changes here. Please refer to other value files under the `.ci/clusters` directory.

.ci/clusters/values-apache-upgrade.yaml
```diff
@@ -0,0 +1,1023 @@
+#
```
sijie, line 1: What do we need a separate `upgrade values file? Can we use the `values-apache-default.yaml` to upgrade?

sijie (commented):
@zymap Can you

zymap:
@sijie PTAL. Thanks

zymap:
Merge this PR first, If has any problem, I will fix it in another PR

PR #278, opened 2021-08-16
Update Pulsar Chart to add APIObject.

MarvinCai:
Update Pulsar Chart to add ZooKeeperCluster, BookKeeperCluster, ProxyCluster and BrokerCluster to indicate controller should manage the cluster.
And add APIObjects to indicate controller should take over existing objects.
Also add knob to disable the chart to create components that controller will create when the cluster will be managed  by controller.

### DIFF
--- a/charts/pulsar/templates/bookkeeper/bookkeeper-cluster.yaml
+++ b/charts/pulsar/templates/bookkeeper/bookkeeper-cluster.yaml
@@ -129,30 +129,31 @@ spec:
         {{- include "pulsar.bookkeeper.ledgers.storage.class" . | nindent 8 }}
     reclaimPolicy: Delete
   {{- end }}
-  conf:
-    zkLedgersRootPath: "{{ .Values.metadataPrefix }}/ledgers"
-    # enable bookkeeper http server
-    httpServerEnabled: "true"
-    httpServerPort: "{{ .Values.bookkeeper.ports.http }}"
-    # config the stats provider
-    statsProviderClass: org.apache.bookkeeper.stats.prometheus.PrometheusMetricsProvider
-    # use hostname as the bookie id
-    useHostNameAsBookieID: "true"
-    {{- if .Values.components.autorecovery }}
-    # disable auto recovery on bookies since we will start AutoRecovery in separated pods
-    autoRecoveryDaemonEnabled: "false"
-    {{- end }}
-    # Do not retain journal files as it increase the disk utilization
-    journalMaxBackups: "0"
-    {{- if .Values.bookkeeper.operator.adopt_existing }}
-    journalDirectories: "/pulsar/data/bookkeeper/journal"
-    PULSAR_PREFIX_journalDirectories: "/pulsar/data/bookkeeper/journal"
-    ledgerDirectories: "/pulsar/data/bookkeeper/ledgers"
-    PULSAR_PREFIX_ledgerDirectories: "/pulsar/data/bookkeeper/ledgers"
-    {{- end}}
-    BOOKIE_MEM: {{ .Values.bookkeeper.configData.BOOKIE_MEM }}
-{{- include "pulsar.bookkeeper.config.tls" . | nindent 4 }}
-{{ (.Files.Glob "conf/bookie/log4j2.yaml").AsConfig | indent 4 }}
+  config:
+    custom:
+      zkLedgersRootPath: "{{ .Values.metadataPrefix }}/ledgers"
+      # enable bookkeeper http server
+      httpServerEnabled: "true"
+      httpServerPort: "{{ .Values.bookkeeper.ports.http }}"
+      # config the stats provider
+      statsProviderClass: org.apache.bookkeeper.stats.prometheus.PrometheusMetricsProvider
+      # use hostname as the bookie id
+      useHostNameAsBookieID: "true"
+      {{- if .Values.components.autorecovery }}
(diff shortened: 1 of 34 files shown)


charts/pulsar/templates/zookeeper/zookeeper-cluster.yaml
```diff
@@ -0,0 +1,223 @@
...
+      storageClassName: {{ .Values.zookeeper.volumes.data.storageClassName }}
+      {{- end -}}
+    {{- end }}
+    reclaimPolicy: Delete
+  {{- end }}
+  config:
+    serverCnxnFactory: org.apache.zookeeper.server.NettyServerCnxnFactory
+    # copy from configmap
+    custom:
+      dataDir: /pulsar/data/zookeeper
```
fantapsody, line 183: Maybe `dataDir` & `dataLogDir` should be configured only when `.Values.zookeeper.operator.adopt_existing` is true?

charts/pulsar/templates/proxy/proxy-cluster.yaml
```diff
@@ -0,0 +1,230 @@
...
+      # prometheus needs to access /metrics endpoint
+      webServicePort: "{{ .Values.proxy.ports.http }}"
+      {{- if or (not .Values.tls.enabled) (not .Values.tls.proxy.enabled) }}
+      servicePort: "{{ .Values.proxy.ports.pulsar }}"
+      {{- end }}
+      {{- if and .Values.tls.enabled .Values.tls.proxy.enabled }}
+      tlsEnabledInProxy: "true"
+      servicePortTls: "{{ .Values.proxy.ports.pulsarssl }}"
+      webServicePortTls: "{{ .Values.proxy.ports.https }}"
+      tlsCertificateFilePath: "/pulsar/certs/proxy/tls.crt"
```
fantapsody, line 159: These cert paths might only be valid when adopting an existing proxy cluster?

charts/pulsar/templates/proxy/proxy-cluster.yaml
```diff
@@ -0,0 +1,230 @@
...
+      extraOptions: []
+      memoryOptions:
+        - {{ .Values.proxy.configData.PULSAR_MEM }}
+      gcOptions:
+        - {{ .Values.proxy.configData.PULSAR_GC }}
+      gcLoggingOptions: []
+  config:
+    tls:
+      enabled: {{ and (.Values.tls.enabled) (.Values.tls.proxy.enabled) }}
+    custom:
```
fantapsody, line 119: We may need to put custom configs in `configData` here?

charts/pulsar/templates/broker/broker-service.yaml
```diff
@@ -27,7 +27,12 @@ metadata:
     {{- include "pulsar.standardLabels" . | nindent 4 }}
     component: {{ .Values.broker.component }}
   annotations:
+    {{- if .Values.broker.service.annotations }}
```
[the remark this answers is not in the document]
MarvinCai: yes, I think in default value file it's just {} (including many other annotations) and causing syntax error, so I added this check.

charts/pulsar/templates/proxy/proxy-cluster.yaml
```diff
@@ -0,0 +1,230 @@
...
+      zookeeperServers: "{{ template "pulsar.zookeeper.connect" . }}{{ .Values.metadataPrefix }}"
```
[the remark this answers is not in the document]
MarvinCai: yeah, you're right, removed.

charts/pulsar/templates/bookkeeper/bookkeeper-cluster.yaml
```diff
@@ -0,0 +1,270 @@
...
+        requests:
+          memory: "{{ .Values.autorecovery.resources.requests.memory }}"
+          cpu: "{{ .Values.autorecovery.resources.requests.cpu }}"
+      {{- end }}
+    conf:
+      {{- include "pulsar.bookkeeper.config.common" . | nindent 6 }}
+{{ toYaml .Values.autorecovery.configData | indent 6 }}
+{{ (.Files.Glob "conf/autorecovery/log4j2.yaml").AsConfig | indent 6 }}
+  apiObjects:
+    clientService: {}
```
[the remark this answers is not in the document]
MarvinCai: actually I don't see where zk/bk client service and broker internal service is used, probably we can disable them as you suggested, will create issue to track that

charts/pulsar/templates/zookeeper/zookeeper-cluster.yaml
MarvinCai: I remember adding the check as you mentioned that to me before, somehow the change was lost, I'll add it here.

MarvinCai: oh nvm, I only add for bk

charts/pulsar/templates/proxy/proxy-cluster.yaml
MarvinCai: that's correct, updated

charts/pulsar/templates/proxy/proxy-cluster.yaml
MarvinCai: .Values.zookeeper.configData is already

fantapsody:
Close this since we have new plans.

PR #303, opened 2021-09-07
fix presto for JDK 11

addisonj:
This fixes the settings for JDK 11... though I realized just now this might break for non JDK 11 images.

Any thoughts on how we should support that?

### DIFF
--- a/charts/pulsar/templates/presto/presto-coordinator-configmap.yaml
+++ b/charts/pulsar/templates/presto/presto-coordinator-configmap.yaml
@@ -35,16 +35,13 @@ data:
     -server
     -Xmx{{ .Values.presto.coordinator.jvm.memory }}
     -XX:+UseG1GC
-    -XX:+UnlockExperimentalVMOptions
-    -XX:+AggressiveOpts
-    -XX:+DoEscapeAnalysis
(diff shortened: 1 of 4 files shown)


sijie (approved):

charts/pulsar/templates/presto/presto-coordinator-configmap.yaml
```diff
@@ -35,16 +35,13 @@ data:
...
-    -XX:+DisableExplicitGC
-    -XX:-ResizePLAB
+    -XX:G1HeapRegionSize=32M
+    -XX:+UseGCOverheadLimit
+    -XX:+ExplicitGCInvokesConcurrent
+    -XX:+HeapDumpOnOutOfMemoryError
     -XX:+ExitOnOutOfMemoryError
-    -XX:+PerfDisableSharedMem
+    -Dpresto-temporarily-allow-java8=true
+    -Djdk.attach.allowAttachSelf=true
```
MarvinCai, line 44: This really limit our ability to use updated Presto/Trino version, current version is 361 and this will limit us to version <= 331 which was release a year and a half ago: https://trino.io/docs/current/release/release-330.html

MarvinCai (approved):

PR #304, opened 2021-09-07
Restore presto service to match service name

addisonj:
Previously, this was changed to be a seperate service, however, this
seems incorrect.

The serviceName should match the name of K8S service which is used for
internal communication.

This effectively reverts #271

@zymap Can you provide some context on why this change was made in the first place?


zymap:
It's a change for a user upgrade from the apache/pulsar-helm-charts. But we use another way to upgrade the user cluster, it's ok to revert this change.

zymap (approved):

PR #307, opened 2021-09-22
Support overriding image pull secrets on broker/proxy/pulsar-manager

sijie:


MarvinCai:
seems we're defining duplicated key, can we do:
```
      {{- if .Values.broker.imagePullSecrets }}
      imagePullSecrets:
      - name: {{ .Values.broker.imagePullSecrets }}
      {{- else }}
      {{- if .Values.imagePullSecrets }}
      imagePullSecrets:
      - name: {{ .Values.imagePullSecrets }}
      {{- end }}
      {{- end }}
```

ericsyh:
Close legacy PR

PR #309, opened 2021-09-23
[MOP] Support TLS-PSK

sijie:
TLS-PSK is introduced since sn-pulsar:2.8.1.1

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -876,6 +876,14 @@ kop:
 mop:
   ports:
     plaintext: 1883
+    tls_psk: 5684
+  # TLS PSK is introduced since sn-pulsar:2.8.1.1
+  tls_psk:
+    enabled: false
+    identity: "[ADD IDENTITY]"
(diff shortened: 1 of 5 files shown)


codelipenghui (approved):

charts/pulsar/values.yaml
```diff
@@ -876,6 +876,14 @@ kop:
 mop:
   ports:
     plaintext: 1883
+    tls_psk: 5684
+  # TLS PSK is introduced since sn-pulsar:2.8.1.1
+  tls_psk:
+    enabled: false
+    identity: "[ADD IDENTITY]"
+    identityHint: "[ADD HINT]"
+    secret: "[ADD SECRET]"
```
Technoboy-, line 885: I‘m a little confused about the config here.
TLS PSK identity in MoP proxy is like : identity1:secret1;identity2:secret2.  It's a semicolon list of string with identity:secret format.

Technoboy-: That means it supports many pairs of the identity.

PR #312, opened 2021-09-28
Add support for AoP.

MarvinCai:


charts/sn-platform/values.yaml
```diff
@@ -934,6 +936,9 @@ broker:
   advertisedPodIP: false
   # For exposing kop service for external access
   advertisedDomain: ""
+  dnsNames:
```
fantapsody, line 939: Would it be better to make it an empty list by default and put those DNS name examples in comments?

charts/sn-platform/templates/broker/broker-cluster.yaml
```diff
@@ -183,6 +187,13 @@ spec:
             key: password
     {{- end }}
     {{- end }}
+    {{- if or .Values.broker.aop.enabled .Values.components.aop }}
+      aop:
+        enabled: true
+        {{- if or .Values.broker.aop.proxyEnabled }}
```
fantapsody, line 193: Is this statement useful?

charts/sn-platform/templates/broker/broker-cluster.yaml
```diff
@@ -183,6 +187,13 @@ spec:
             key: password
     {{- end }}
     {{- end }}
+    {{- if or .Values.broker.aop.enabled .Values.components.aop }}
```
fantapsody, line 190: It's a bit confusing to have multiple flags to enable a feature...

charts/sn-platform/values.yaml
```diff
@@ -945,6 +950,9 @@ broker:
       #  key: password
       #  name: kop-keystore-password
   # choose istio selector
+  aop:
+    enabled: false
+    proxyEnabled: false
```
fantapsody, line 955: Would users prefer the proxy to be enabled by default?

fantapsody (approved):
Overall looks good, left some comments.

charts/sn-platform/values.yaml
MarvinCai: sure, will comment it out.

charts/sn-platform/templates/broker/broker-cluster.yaml
MarvinCai: honestly yes, but I'm not sure why we have 2 different flags, I assume user will expect kop/aop enabled if we set either of them so just check both here.

charts/sn-platform/values.yaml
MarvinCai: I think value's default is for most user who might not need aop enabled, so keep it to false by default.

charts/sn-platform/templates/broker/broker-cluster.yaml
MarvinCai: removed the or

PR #313, opened 2021-09-29
Fixed grafana root url

tuteng:
Document https://grafana.com/docs/grafana/latest/administration/configuration/#root_url
This is the full URL used to access Grafana from a web browser. This is important if you use Google or GitHub OAuth authentication (for the callback URL to be correct).

### DIFF
--- a/charts/sn-platform/templates/grafana/grafana-deployment.yaml
+++ b/charts/sn-platform/templates/grafana/grafana-deployment.yaml
@@ -97,7 +97,7 @@ spec:
         - name: GRAFANA_DOMAIN
           value: {{ template "pulsar.control_center_domain" . }}
         - name: GRAFANA_ROOT_URL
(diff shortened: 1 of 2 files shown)


charts/sn-platform/templates/grafana/grafana-deployment.yaml
```diff
@@ -97,7 +97,7 @@ spec:
         - name: GRAFANA_DOMAIN
           value: {{ template "pulsar.control_center_domain" . }}
         - name: GRAFANA_ROOT_URL
-          value: {{ template "pulsar.control_center_url" . }}{{ template "pulsar.control_center_path.grafana" . }}/
+          value: https://{{ template "pulsar.control_center_domain" . }}{{ template "pulsar.control_center_url" . }}{{ template "pulsar.control_center_path.grafana" . }}/
```
tuteng, line 100: I'm not sure if this schema should use `https`, if the customer uses ingress to expose a public host, ideally it should enable https

sijie (approved):

PR #317, opened 2021-10-04
Fixed custom grafana configuration file

tuteng:
Reference https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L545
* Fixed custom grafana configuration file
* Add Azure AD configuration

### DIFF
--- a/charts/sn-platform/values.yaml
+++ b/charts/sn-platform/values.yaml
@@ -1487,6 +1487,38 @@ datadog:
 ##
 grafana:
   component: grafana
+  grafana.ini:
+    paths:
+      data: /var/lib/grafana/pulsar/data
+      plugins: /var/lib/grafana/pulsar/plugins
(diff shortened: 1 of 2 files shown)


sijie (approved):

charts/sn-platform/values.yaml
```diff
@@ -1487,6 +1487,38 @@ datadog:
...
+      name: Azure AD
+      enabled: false
+      allow_sign_up: true
+      client_id: ""
+      client_secret: ""
+      scopes: openid email profile
+      auth_url: ""
+      token_url: ""
+      allowed_domains: ""
+      allowed_groups: ""
```
EronWright, line 1514: Make it possible to set `role_attribute_strict = true` to map users to Grafana roles based on Azure AD App Roles.

PR #358, opened 2021-10-28
[sn-console] Support Google login

sijie:


Huanli-Meng:
The doc is updated for this PR. Therefore, I remove doc-required label and add doc-added label.

PR #362, opened 2021-10-29
Add filesystem offloader configurations

zymap:
---

*Motivation*

Allow enabling the filesystem offloader with helm charts.

*Modifications*

Add filesystem offloader configurations in the pulsar charts.

### DIFF
--- a/charts/pulsar/templates/broker/_broker.tpl
+++ b/charts/pulsar/templates/broker/_broker.tpl
@@ -350,6 +350,24 @@ Define gcs offload options mounts
 {{- end }}
 {{- end }}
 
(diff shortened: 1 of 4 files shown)


MarvinCai:
shall we also update sn-platform chart?

sijie (approved):

sijie:
> shall we also update sn-platform chart?

Yes. We should do that same thing. @zymap can you update the sn-platform as well?

zymap:
Sure. I opened an [issue](https://github.com/streamnative/charts/issues/382) and will do it in another PR.

PR #368, opened 2021-10-29
Fixed google oauth2 login failed bug

tuteng:
* Set default value is false
* Add google oauth2 secret file


sijie (approved):

Huanli-Meng:
The doc is updated. Therefore, i remove the doc-required label and add doc-added label.

PR #373, opened 2021-11-01
Update pulsar/sn-platform chart to use rbacv1 api.

MarvinCai:
Should be safe as rbac.authorization.k8s.io/v1 is available isnce k8s v1.8

### DIFF
--- a/charts/pulsar/templates/bookkeeper/bookkeeper-cluster-role-binding.yaml
+++ b/charts/pulsar/templates/bookkeeper/bookkeeper-cluster-role-binding.yaml
@@ -35,7 +35,7 @@ subjects:
   namespace: {{ template "pulsar.namespace" . }}
 ---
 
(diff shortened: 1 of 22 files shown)


fantapsody (approved):

tuteng (commented):
Trying to upgrade encountered the following exception, not sure if this is expected:
```
Error: UPGRADE FAILED: rendered manifests contain a new resource that already exists. Unable to continue with update: existing resource conflict: kind: ClusterRole, namespace: , name: pulsar2-sn-platform-bookie-clusterrole
```

tuteng (approved):

MarvinCai:
synced offline, after updating the helm from 3.0.1 to a newer version the upgrade just work, also k8s will auto convert rbac.v1beta1 resources to rbac.v1 resources so user shoouldn't need a upgrade, but this pr will just prevent error for fresh install when installing on k8s 1.22+ clusters.

PR #377, opened 2021-11-05
[sn-console] Support configuring redirect host

sijie:


tuteng (approved):

charts/sn-console/templates/console/streamnative-console-statefulset.yaml
```diff
@@ -118,6 +118,12 @@ spec:
           - name: REDIRECT_PORT
             value: "{{ .Values.streamnative_console.configData.REDIRECT_PORT }}"
           {{- end }}
```
tuteng, line 120: It's fixed here https://github.com/streamnative/charts/blob/master/charts/sn-platform/templates/streamnative-console/streamnative-console-statefulset.yaml#L147, do we still need this pr

PR #385, opened 2021-11-10
Add default org and instance name for console

tuteng:
When starting platform, some users often forget to configure the organization and instance name, so add a default value here, and you can override it in your custom values.yaml file


ericsyh (approved):
LGTM

PR #388, opened 2021-11-17
fix(charts): correct the pulsar manager image version

ericsyh:
Signed-off-by: Eric Shen <ericshenyuhao@outlook.com>

### Motivation

We don't have the 0.6.0 image tag for [streamnative/pulsar-manager](https://hub.docker.com/r/streamnative/pulsar-manager/tags), so correct the tag to 0.3.0

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -176,7 +176,7 @@ images:
     pullPolicy: IfNotPresent
   pulsar_manager:
     repository: streamnative/pulsar-manager
-    tag: 0.6.0
+    tag: 0.3.0
     pullPolicy: IfNotPresent
     hasCommand: false
   node_exporter:


nlu90 (commented):
0.3.0 is pushed two years ago...

Are you sure this is the desired version we want to use? Also, by any chance, should we publish new pulsar-manager images?

ericsyh:
@nlu90 My intention for this fix is to provide a correct chart for users, in case they intall the chart and find no target pulsar-manager image provided. 
I think i can check with @tuteng whether we can make new release for pulsar manager since the current version is quite long time ago.

nlu90 (approved):

tuteng (approved):

tuteng:
A new release will be considered recently

tuteng:
> @nlu90 My intention for this fix is to provide a correct chart for users, in case they intall the chart and find no target pulsar-manager image provided. I think i can check with @tuteng whether we can make new release for pulsar manager since the current version is quite long time ago.

Version 0.6.0 seems to be from sn-pulsar-manager https://hub.docker.com/r/streamnative/sn-pulsar-manager/tags

PR #390, opened 2021-11-18
Add global resource global read only option

tuteng:
* Add global resource global read only option


urfreespace (approved):

Huanli-Meng:
The doc is updated in this [PR](https://github.com/streamnative/doc-platform/pull/384).  Therefore, remove the doc-required label and add doc-added label.

PR #400, opened 2021-12-02
Unify all image version to one place

tuteng:
* Unify all image version to one place

- [x] Do test


MarvinCai (approved):
lgtm

sijie (approved):
LGTM

charts/sn-platform/values.yaml
```diff
@@ -208,6 +219,59 @@ images:
...
+    tag: v1.3.1
+    pullPolicy: IfNotPresent
+  cert_manager_webhook:
+    repository: quay.io/jetstack/cert-manager-controller
+    tag: v1.3.1
+    pullPolicy: IfNotPresent
+  vault_operator:
+    repository: ghcr.io/banzaicloud/vault-operator
+    tag: v1.3.0
+    pullPolicy: IfNotPresent
```
ericsyh, line 274: @tuteng Adding images for cert-manager, pulsar-operator and function-mesh in this values doesn't appear to work. There is no template related in sn-platform chart.

tuteng: Yes, this section just puts all the image needed in snp here, so that users can easily find all the image used in snp

ericsyh: I see, but it will be quite confusing for users. I think it will be better if we put the images list on Platform docs.

PR #403, opened 2021-12-03
Fix issue that bookkeeper don't support some jvm options anymore on java11

FushuWang:
The image version of pulsar has upgraded to v2.8.x, there are a few JVM options that the bookkeeper doesn't support anymore.
It should be removed from the `vaules.yaml` as the default configuration

```
configData:
    PULSAR_GC: >
      ....
      -XX:+PrintGCTimeStamps
      -XX:+PrintGCApplicationStoppedTime
      -XX:+PrintHeapAtGC
      -XX:G1LogLevel=finest
      ....
```


fantapsody (approved):

MarvinCai:
It might be fine to remove them, but I believe we add such config before for good reason, so I think we should consider adding equivalent in java11.

MarvinCai:
or we should make it configurable, as some user might want to stick with old pulsar version

tuteng:
> or we should make it configurable, as some user might want to stick with old pulsar version

+1, Perhaps a more appropriate approach would be to comment out these options and add some introductions

FushuWang:
According to these links and warnings of bookkeeper
[Unrecognized JVM GC options when using JAVA 11](https://confluence.atlassian.com/confkb/unrecognized-jvm-gc-options-when-using-java-11-1002472841.html)
[stackoverflow](https://stackoverflow.com/questions/54144713/is-there-a-replacement-for-the-garbage-collection-jvm-args-in-java-11)
[Convert GC Logging Flags to Xlog](https://docs.oracle.com/javase/9/tools/java.htm#GUID-BE93ABDC-999C-4CB5-A88B-1994AAAC74D5__CONVERTGCLOGGINGFLAGSTOXLOG-A5046BD1)

```shell
Skipping rack-awareness setup
OpenJDK 64-Bit Server VM warning: Option AggressiveOpts was deprecated in version 11.0 and will likely be removed in a future release.
[0.002s][warning][gc] -Xloggc is deprecated. Will use -Xlog:gc:/var/log/bookie-gc.log instead.
[0.003s][warning][gc] -XX:+PrintGCDetails is deprecated. Will use -Xlog:gc* instead.
```

There are the replacements for these options:
|VM option in Java8| The replacement in Java11|
|:-:|:-|
| -XX:+PrintGCDetails | -Xlog:gc* |
| -XX:+PrintGCTimeStamps | -Xlog:::time,level,tags |

tuteng (approved):

sijie (approved):

PR #404, opened 2021-12-04
Add resources to support auto scaling.

MarvinCai:


charts/sn-platform/templates/hpa/hpa-init.yaml
```diff
@@ -0,0 +1,140 @@
...
+    metadata:
+      name: ${CERT_NAME}
+      namespace: ${CERT_NAMESPACE}
+    data:
+      serving.crt: $(cat apiserver.pem | base64 ${b64_opts})
+      serving.key: $(cat apiserver-key.pem | base64 ${b64_opts})
+    EOF
+---
+# The job basically execute above script to generate a cert for kube-aggregator to communicate with metric server.
+apiVersion: batch/v1
```
FushuWang, line 101: Do there need  the control of `.Values.initialize`, I saw there are some init jobs under the control of it

charts/sn-platform/values.yaml
```diff
@@ -28,7 +28,7 @@ initialize: false
 # NOTE: Make the default namespace as empty. So it will fallback to use the namespace used for installing the helm
 #       chart. Helm does not position it self as a namespace manager, as namespaces in kubernetes are considered as
 #       a higher control structure that is not part of the application.
-namespace: ""
+namespace: "zxc"
```
FushuWang, line 31: need to remove it

MarvinCai: oops, had to add namespace for testing, will remove it.

charts/sn-platform/templates/hpa/hpa-init.yaml
MarvinCai: good question, after some thought, this one is a bit different, it's idempotent(end result is a secret with self signed cert for custom metric server but might with different expiration date), and it can be run multiple times, unlike other job might create meta data in zk or bk which can't be execute multiple times
also users who already has the chart installed might still want to enable hpa, but likely they can't the initialize back to false to trigger that,
so I think it's fine to not use .Values.initialize to control this job

fantapsody (approved):

PR #412, opened 2021-12-11
Workaround for CVE-2021-44228 Log4J RCE when Log4J >= 2.10.0.

MarvinCai:
Presto is fine as it use its own airlift logging library.

### DIFF
--- a/charts/pulsar/templates/bookkeeper/bookkeeper-autorecovery-statefulset.yaml
+++ b/charts/pulsar/templates/bookkeeper/bookkeeper-autorecovery-statefulset.yaml
@@ -145,7 +145,7 @@ spec:
         - >
           bin/apply-config-from-env.py conf/bookkeeper.conf;
           {{- include "pulsar.autorecovery.zookeeper.tls.settings" . | nindent 10 }}
-          OPTS="${OPTS} -Dlog4j2.formatMsgNoLookups=true" bin/bookkeeper autorecovery
+          bin/bookkeeper autorecovery
         ports:
         - name: http
           containerPort: {{ .Values.autorecovery.ports.http }}
--- a/charts/pulsar/templates/bookkeeper/bookkeeper-autorecovery-statefulset.yaml
+++ b/charts/pulsar/templates/bookkeeper/bookkeeper-autorecovery-statefulset.yaml
@@ -145,7 +145,7 @@ spec:
         - >
           bin/apply-config-from-env.py conf/bookkeeper.conf;
           {{- include "pulsar.autorecovery.zookeeper.tls.settings" . | nindent 10 }}
-          bin/bookkeeper autorecovery
+          OPTS="${OPTS} -Dlog4j2.formatMsgNoLookups=true" bin/bookkeeper autorecovery
         ports:
         - name: http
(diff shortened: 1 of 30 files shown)


MarvinCai:
I’m doing some more test.

fantapsody (approved):

charts/sn-platform/templates/bookkeeper/bookkeeper-cluster.yaml
```diff
@@ -107,6 +107,8 @@ spec:
         - {{ .Values.bookkeeper.configData.PULSAR_MEM }}
       gcOptions:
         - {{ .Values.bookkeeper.configData.PULSAR_GC }}
+      extraOptions:
```
EronWright, line 110: @fantapsody is this an example of the situation we talked about, that the operator applies various default values only when `extraOptions` is `[]`.  If so, then this change might cause other options to change, because they don't be applied.

EronWright (changes requested):

charts/sn-platform/templates/bookkeeper/bookkeeper-cluster.yaml
MarvinCai: @EronWright seems so, which is not reasonable IMO, user provided ones should able to override whatever existing config, I think we should update the operator to put user provided arg at the end of jvm args as seems the right-most arg always wins
for this change I think we can just explicitly add whatever default jvm args operator will apply so we'll still have them, WDYT

EronWright: Yes please do that.

charts/pulsar/templates/bookkeeper/bookkeeper-autorecovery-statefulset.yaml
```diff
@@ -145,7 +145,7 @@ spec:
         - >
           bin/apply-config-from-env.py conf/bookkeeper.conf;
           {{- include "pulsar.autorecovery.zookeeper.tls.settings" . | nindent 10 }}
-          bin/bookkeeper autorecovery
+          OPTS="${OPTS} -Dlog4j2.formatMsgNoLookups=true" bin/bookkeeper autorecovery
```
sijie, line 148: @MarvinCai Can you use `PULSAR_EXTRA_OPTS` to pass the flag in? You can add it into configmap instead of modifying the commandline.

sijie (changes requested):

charts/sn-platform/templates/bookkeeper/bookkeeper-cluster.yaml
MarvinCai: updated, PTAL

charts/pulsar/templates/bookkeeper/bookkeeper-autorecovery-statefulset.yaml
MarvinCai: @sijie updated, PTAL

sijie (approved):

sijie:
@MarvinCai Can you rebase the PR and resolve the CI issues?

Issue #423, opened 2021-12-20
authentication params for pulsar-detector


ignacioli:
# Motivation 

The current helm charts for pulsar-detector has no auth configuration segment, we need to add them.

I'v checked the former issue #92 , but I dont know why it's closed in an unsolved status, and this issue: #409 , I think it's not a graceful way, may lead potential complication.

# Request
Add authentication params for pulsar-detector configuration segment by default.

ignacioli:
do not need to do that, will close it.

PR #424, opened 2021-12-20
Add auth parameter for detector and enable auth by default

ignacioli:
#423

### DIFF
--- a/charts/pulsar/templates/detector/pulsar-detector-statefulset.yaml
+++ b/charts/pulsar/templates/detector/pulsar-detector-statefulset.yaml
@@ -85,8 +85,19 @@ spec:
         args:
         - >
           bin/pulsar-detector -service-url pulsar://{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}:{{ .Values.broker.ports.pulsar }};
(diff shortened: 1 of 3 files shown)


ignacioli:
pulsar_detector image : keep same as charts sn-platform.

sijie:
@ignacioli Can you resolve the conflict?

charts/pulsar/templates/detector/pulsar-detector-statefulset.yaml
```diff
@@ -85,8 +85,19 @@ spec:
         args:
         - >
           bin/pulsar-detector -service-url pulsar://{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}:{{ .Values.broker.ports.pulsar }};
```
sijie, line 87: Don't we need to delete this line?

charts/pulsar/values.yaml
```diff
@@ -95,7 +95,7 @@ components:
   # mop
   mop: false
   # pulsar detector
-  pulsar_detector: false
+  pulsar_detector: true
```
sijie, line 98: I don't think we should turn this one by default.

charts/pulsar/values.yaml
```diff
@@ -249,7 +249,7 @@ tls:
 # Enable or disable broker authentication and authorization.
 auth:
   authentication:
-    enabled: false
+    enabled: true
```
sijie, line 252: Same comment. We shouldn't turn this one by default

sijie (changes requested):

Issue #420, opened 2021-12-16
Run Function Worker as a separate statefulset


sijie:
*Motivation*

Currently, the helm chart runs the Function Worker as part of brokers. It has a stability issue. 

It would be good to move the Function Worker to a separate statefulset.

*Modification*

1. Add a statefulset for Function Worker.
2. Configure the proxy to talk to the new StatefulSet of Function Worker.
```
# If function workers are setup in a separate cluster, configure the following 2 settings
# to point to the function workers cluster
functionWorkerWebServiceURL=
functionWorkerWebServiceURLTLS=
```
3. Disable `functionsWorkerEnabled` in broker.

PR #426, opened 2021-12-20
Create dedicate statefulset to run function worker if needed.

MarvinCai:
Fixes https://github.com/streamnative/charts/issues/420


charts/pulsar/values.yaml
```diff
@@ -1051,6 +1056,8 @@ broker:
 ##
 functions:
   component: functions-worker
+  useDedicateRunner: true
```
nlu90, line 1059: should we set to `false` be default?

charts/pulsar/templates/function-worker/function-worker-statefulset.yaml
```diff
@@ -0,0 +1,204 @@
...
+                        operator: In
+                        values:
+                          - {{ .Values.functions.component }}
+                  topologyKey: "kubernetes.io/hostname"
+        {{ end }}
+      {{- end }}
+      terminationGracePeriodSeconds: {{ .Values.broker.gracePeriod }}
+      initContainers:
+        # This init container will wait for zookeeper to be ready before
+        # deploying the bookies
```
nlu90, line 119: misleading comments

charts/pulsar/templates/function-worker/function-worker-statefulset.yaml
```diff
@@ -0,0 +1,204 @@
...
+        - name: wait-zookeeper-ready
```
nlu90, line 120: function worker doesn't talk to zk directly, so it doesn't need to wait for zk ready IMO.

charts/pulsar/templates/function-worker/function-worker-role-binding.yaml
```diff
@@ -0,0 +1,68 @@
...
+    {{- include "pulsar.standardLabels" . | nindent 4 }}
+rules:
+- apiGroups: [""]
+  resources:
+  - configmap
+  - configmaps
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["", "extensions", "apps"]
+  resources:
+    - pods
```
nlu90, line 52: does it need access to `configmap?`

charts/pulsar/values.yaml
MarvinCai: my bad, changed to true for testing, updated

charts/pulsar/templates/function-worker/function-worker-statefulset.yaml
MarvinCai: the init containers are actually copied from proxy, I assume proxy and function worker both depend on broker so should have same prerequisite, if function worker doesn't need to wait for zk then I'll remove this one.

charts/pulsar/templates/function-worker/function-worker-statefulset.yaml
MarvinCai: see below comment.

freeznet (approved):

charts/pulsar/templates/function-worker/function-worker-role-binding.yaml
MarvinCai: seems we don't need configmap permission, removed.

MarvinCai: somehow seems the document said configmap permission is needed, I think better just keep it: https://pulsar.apache.org/docs/en/functions-runtime/

charts/pulsar/templates/function-worker/function-worker-statefulset.yaml
```diff
@@ -0,0 +1,188 @@
...
+          {{- include "pulsar.function.worker.config.volumeMounts" . | nindent 10 }}
+{{- with .Values.functions.extraVolumeMounts }}
+{{ toYaml . | indent 10 }}
+{{- end }}
+      volumes:
+      {{- include "pulsar.function.worker.config.volumes" . | nindent 6 }}
+{{- with .Values.functions.extraVolumes }}
+{{ toYaml . | indent 6 }}
+{{- end }}
+  {{- end }}
```
fantapsody, line 188: Seems two extra white spaces here?

charts/pulsar/templates/function-worker/function-worker-statefulset.yaml
```diff
@@ -0,0 +1,188 @@
...
+                        operator: In
+                        values:
+                          - {{ .Values.functions.component }}
+                  topologyKey: "kubernetes.io/hostname"
+        {{ end }}
+      {{- end }}
+      terminationGracePeriodSeconds: {{ .Values.broker.gracePeriod }}
+      initContainers:
+        # This init container will wait for pulsar broker to be ready before deploying the function worker
+        - name: wait-proxy-ready
```
fantapsody, line 119: Why does the function worker need to wait for the proxy?

charts/pulsar/templates/broker/broker-statefulset.yaml
```diff
@@ -62,7 +62,7 @@ spec:
 {{- with .Values.broker.securityContext }}
 {{ toYaml . | indent 8 }}
 {{- end }}
-    {{- if or .Values.broker.serviceAccount.use .Values.components.functions }}
+    {{- if or .Values.broker.serviceAccount.use (and .Values.components.functions  (not .Values.functions.useDedicateRunner)) }}
```
fantapsody, line 65: The change looks to be equivalent to the original code. I'm just curious if there are other use cases that also requires a dedicated service account for brokers, like offloading data to the object storage with some workload identities.

charts/pulsar/templates/function-worker/_function_worker.tpl
```diff
@@ -0,0 +1,152 @@
...
+{{- end }}
+
+
+{{/*
+Define function token mounts
+*/}}
+{{- define "pulsar.function.token.volumeMounts" -}}
+{{- if .Values.auth.authentication.enabled }}
+{{- if eq .Values.auth.authentication.provider "jwt" }}
+{{- if not .Values.auth.vault.enabled }}
```
tuteng, line 108: This chart does not support vault authentication, and it seems that you can remove this check

tuteng (approved):
Did some deployment tests in the test environment and it all seems to work

charts/pulsar/values.yaml
MarvinCai: done

charts/pulsar/templates/function-worker/function-worker-statefulset.yaml
MarvinCai: this won't cause syntax error, but fixed

charts/pulsar/templates/function-worker/function-worker-statefulset.yaml
MarvinCai: seems the function worker talk to proxy instead of directly talk to broker. I've tried broker endpoint which doesn't work.

charts/pulsar/templates/broker/broker-statefulset.yaml
MarvinCai: yeah, offloading could be the other use case, but for now just trying to keep the same behavior as before.

charts/pulsar/templates/function-worker/function-worker-role-binding.yaml
sijie: I believe we just need `get`, `list`, and `watch` permission on configmap. But that should have been covered by https://github.com/streamnative/charts/pull/426/files#diff-437f94e036fc125552f2cb0f10309621cbd593de448938ff6bfbb332964cc6a8R45

sijie (approved):

Huanli-Meng:
as comment by Eric in doc-platform repo, https://github.com/streamnative/charts/issues/575, we will set Function Mesh + unction Mesh Worker Service by default on SNP 1.5, so there is no need to mention different function worker deployment modes on docs.

PR #433, opened 2021-12-28
PulsarProxy external service support type ClusterIP/NodePort

FushuWang:
Update the crd of PulsrProxy.  For this [PR](https://github.com/streamnative/pulsar-operators/pull/469)

### DIFF
--- a/charts/sn-platform/templates/proxy/proxy-cluster.yaml
+++ b/charts/sn-platform/templates/proxy/proxy-cluster.yaml
@@ -189,10 +189,17 @@ spec:
   {{- if .Values.ingress.proxy.enabled }}
       metadata:
         name: "{{ template "pulsar.fullname" . }}-{{ .Values.proxy.component }}-ingress"
+        {{- if .Values.ingress.proxy.annotations }}
         annotations:
         {{- with .Values.ingress.proxy.annotations }}
           {{- toYaml . | nindent 10 }}
         {{- end }}
+        {{- end }}
+      type: {{ .Values.ingress.proxy.type }}
+      {{- if and (eq .Values.ingress.proxy.type "LoadBalancer") .Values.ingress.proxy.exposeSSLPort (not .Values.ingress.proxy.tls.enabled) }}
+      exposeSSLPort: {{ .Values.ingress.proxy.exposeSSLPort }}
+      {{- end }}
+
   {{- else }}
       {}
   {{- end }}
--- a/charts/pulsar-operator/crds/pulsar.streamnative.io_pulsarproxies.yaml
+++ b/charts/pulsar-operator/crds/pulsar.streamnative.io_pulsarproxies.yaml
@@ -117,6 +117,7 @@ spec:
                               description: Number or name of the port to access on
                                 the pods targeted by the service. When add new service
                                 port, it shouldn't be empty
+                              format: int32
                               type: integer
                           type: object
                         type: array
--- a/charts/pulsar-operator/crds/pulsar.streamnative.io_pulsarproxies.yaml
+++ b/charts/pulsar-operator/crds/pulsar.streamnative.io_pulsarproxies.yaml
@@ -11,231 +11,272 @@ spec:
   group: pulsar.streamnative.io
   names:
     categories:
-      - pulsar
+    - pulsar
     kind: PulsarProxy
     listKind: PulsarProxyList
     plural: pulsarproxies
     shortNames:
-      - pp
-      - proxy
+    - pp
+    - proxy
     singular: pulsarproxy
   scope: Namespaced
   versions:
-    - additionalPrinterColumns:
-        - jsonPath: .spec.replicas
-          name: Replicas
-          type: integer
-        - jsonPath: .status.replicas
-          name: Ready Replicas
-          type: integer
-        - jsonPath: .spec.image
-          name: Desired Image
-          type: string
-        - jsonPath: .spec.dnsNames[0]
-          name: Endpoint
-          type: integer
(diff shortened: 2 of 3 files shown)


fantapsody (commented):
1. could you add some context in the description?
2. could you make it be compatible with the proxy ingress settings defined in https://github.com/streamnative/charts/blob/master/charts/sn-platform/templates/proxy/proxy-service-ingress.yaml#L39?
  - for example, it seems `exposeSSLPort` should be `true` if `.Values.ingress.proxy.tls.enabled` is `true` but `.Values.ingress.proxy.tls.enabled` is `false` if `exposeSSLPort` is not defined?

FushuWang:
> 1. could you add some context in the description?
> 2. could you make it be compatible with the proxy ingress settings defined in https://github.com/streamnative/charts/blob/master/charts/sn-platform/templates/proxy/proxy-service-ingress.yaml#L39?
> 
> * for example, it seems `exposeSSLPort` should be `true` if `.Values.ingress.proxy.tls.enabled` is `true` but `.Values.ingress.proxy.tls.enabled` is `false` if `exposeSSLPort` is not defined?

OK

FushuWang:
> 1. could you add some context in the description?
> 2. could you make it be compatible with the proxy ingress settings defined in https://github.com/streamnative/charts/blob/master/charts/sn-platform/templates/proxy/proxy-service-ingress.yaml#L39?
> 
> * for example, it seems `exposeSSLPort` should be `true` if `.Values.ingress.proxy.tls.enabled` is `true` but `.Values.ingress.proxy.tls.enabled` is `false` if `exposeSSLPort` is not defined?

The `exposeSSLPort` only works when `.Values.ingress.proxy.tls.enabled`  is `false`, if `.Values.ingress.proxy.tls.enabled` is `true`, it will keep the same behavior with before.

MarvinCai:
> 1. could you add some context in the description?
> 2. could you make it be compatible with the proxy ingress settings defined in https://github.com/streamnative/charts/blob/master/charts/sn-platform/templates/proxy/proxy-service-ingress.yaml#L39?
> 
> * for example, it seems `exposeSSLPort` should be `true` if `.Values.ingress.proxy.tls.enabled` is `true` but `.Values.ingress.proxy.tls.enabled` is `false` if `exposeSSLPort` is not defined?

I think if `.Values.ingress.proxy.tls.enabled` is `true` then `exposeSSLPort` should be `false` , as enable tls in proxy means we want to terminate tls traffic at proxy and ingress service should also open tls port and forward to proxy tls port, but `exposeSSLPort` = true means proxy doesn't open tls port, only ingress will open tls port and ingress itself will terminate tls traffic.
So the only situation is `.Values.ingress.proxy.tls.enabled`=`false` and `exposeSSLPort`= `true`
Does that make sense?

fantapsody:
> > 1. could you add some context in the description?
> > 2. could you make it be compatible with the proxy ingress settings defined in https://github.com/streamnative/charts/blob/master/charts/sn-platform/templates/proxy/proxy-service-ingress.yaml#L39?
> > 
> > 
> > * for example, it seems `exposeSSLPort` should be `true` if `.Values.ingress.proxy.tls.enabled` is `true` but `.Values.ingress.proxy.tls.enabled` is `false` if `exposeSSLPort` is not defined?
> 
> I think if `.Values.ingress.proxy.tls.enabled` is `true` then `exposeSSLPort` should be `false` , as enable tls in proxy means we want to terminate tls traffic at proxy and ingress service should also open tls port and forward to proxy tls port, but `exposeSSLPort` = true means proxy doesn't open tls port, only ingress will open tls port and ingress itself will terminate tls traffic. So the only situation is `.Values.ingress.proxy.tls.enabled`=`false` and `exposeSSLPort`= `true` Does that make sense?

Sorry, I actually meant the `exposeSSLPort` should be `true` when `.Values.ingress.proxy.tls.enabled` is true and `and (.Values.tls.enabled) (.Values.tls.proxy.enabled)` is `false`, as the former indicates whether the users want to expose the pulsar proxy service on TLS or not, and the latter indicates whether the `TLS` will be enabled in the proxy service directly. Is that what you thought? @FushuWang @MarvinCai

MarvinCai:
I think these 2 flags might just cause confusion as tls setting should be consistent for ingress service and pulsar proxy (without this exposeSSLPort tweak), else the port mapping just won't match, so I think 1 flag proxy.tls should be just enough, but that's out of scope of this pr.
I think your suggestion make sense.

charts/sn-platform/templates/proxy/proxy-cluster.yaml
```diff
@@ -189,10 +189,21 @@ spec:
...
+        {{- if .Values.ingress.proxy.annotations }}
         annotations:
         {{- with .Values.ingress.proxy.annotations }}
           {{- toYaml . | nindent 10 }}
         {{- end }}
+        {{- end }}
+      {{- if ne .Values.ingress.proxy.type "LoadBalancer" }}
+      type: {{ .Values.ingress.proxy.type }}
+      {{ else }}
+      {{- if and .Values.ingress.proxy.tls.enabled (not .Values.tls.enabled) (not .Values.tls.enabled)}}
```
fantapsody, line 201: Why two consecutive `(not .Values.tls.enabled)` here?

fantapsody (changes requested):

charts/sn-platform/templates/proxy/proxy-cluster.yaml
FushuWang: ops, the second should be `(not .Values.tls.proxy.enabled)`.

fantapsody (approved):

PR #437, opened 2022-01-07
feat: oauth2 for AzureAd optimize

urfreespace:


maxsxu (commented):
@ericsyh @urfreespace Do we still need this or procceed?

PR #440, opened 2022-01-12
Use separate image for function-worker.

MarvinCai:


sijie (approved):

Huanli-Meng:
as comment by Eric in `doc-platform` repo, https://github.com/streamnative/charts/issues/575, we will set Function Mesh + unction Mesh Worker Service by default on SNP 1.5, so there is no need to mention different function worker deployment modes on docs.

PR #442, opened 2022-01-12
add support for custom affinity rules

addisonj:
This commit add support for custom affinity rules to be more selective

### DIFF
--- a/charts/pulsar/templates/bookkeeper/bookkeeper-autorecovery-statefulset.yaml
+++ b/charts/pulsar/templates/bookkeeper/bookkeeper-autorecovery-statefulset.yaml
@@ -70,6 +70,10 @@ spec:
 {{ toYaml . | indent 8 }}
 {{- end }}
     {{- end }}
+      {{- if and .Values.affinity.anti_affinity .Values.autorecovery.custom_affinity}}
+      affinity:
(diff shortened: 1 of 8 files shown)


charts/pulsar/templates/bookkeeper/bookkeeper-autorecovery-statefulset.yaml
```diff
@@ -70,6 +70,10 @@ spec:
 {{ toYaml . | indent 8 }}
 {{- end }}
     {{- end }}
+      {{- if and .Values.affinity.anti_affinity .Values.autorecovery.custom_affinity}}
```
fantapsody, line 73: Overall looks good. The flags used here seems a bit confusing when I took the first look at them.

fantapsody (approved):

charts/pulsar/templates/bookkeeper/bookkeeper-autorecovery-statefulset.yaml
addisonj: yes... I agree, but it is the pattern already used so we are a bit locked in here :\

MarvinCai (approved):

addisonj:
This was added in https://github.com/streamnative/charts/pull/478

PR #446, opened 2022-01-24
Disable authentication for KoP by default.

MarvinCai:
In KoP doc, auth is disabled by default, so if we enable auth for KoP by default might cause
surprise for user following KoP doc.
ref: https://github.com/streamnative/kop/blob/master/docs/security.md


fantapsody (approved):

FushuWang (approved):
LGTM

PR #449, opened 2022-01-29
Dedicated function worker only enable auth and broker client auth when broker enabled auth.

MarvinCai:
resolve #449


fantapsody (approved):

tuteng (approved):

Huanli-Meng:
https://github.com/streamnative/charts/issues/575, we will set Function Mesh + unction Mesh Worker Service by default on SNP 1.5, so there is no need to mention different function worker deployment modes on docs.

PR #464, opened 2022-02-01
Add a new vault job for init public key file

tuteng:
* Add a new vault job for init public key file


FushuWang (approved):
LGTM, and I test the upgrade from 1.2.30, it's okay.

PR #447, opened 2022-01-26
Upgrade deprecated Ingress apiVersion extensions/v1beta1 to networking.k8s.io/v1

FushuWang:
Resolve https://github.com/streamnative/data-plane-epics/issues/1

When installing sn-platform by the chart, it will get these warnings
```shell
W0126 15:01:15.043974   85780 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
W0126 15:01:15.213499   85780 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
W0126 15:01:24.922246   85780 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0126 15:01:47.110670   85780 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
W0126 15:01:47.197697   85780 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
W0126 15:01:51.309045   85780 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```

Especially, Ingress apiVersion `extensions/v1beta1` will be unavailable in v1.22+.  

This PR will not fix the warning about PodDisruptionBudget。

For chart compatibility, when Kubernetes version is less than v1.19, it will still use `extensions/v1beta1` for `Ingress`, when it's greater or equal to v1.19, it will use `networking.k8s.io/v1`
| kind |  apiVersion | kubeVersion |
| :-: | :-: | :-: |
| Ingress | `extensions/v1beta1` | < v1.19 |
| Ingress | `networking.k8s.io/v1` | >= v1.19 |

## Test cases
- install it into k8s v1.21 and v1.22
- install it into k8s v1.18
- upgrade from sn-platform:1.2.30 in k8s v1.18
- upgrade from sn-platform:1.2.30 in k8s v1.21

PR #465, opened 2022-02-01
[sn-platform]Upgrade nginx ingress controller to v1.1.1 for k8s v1.22+

FushuWang:
# Motivation
To support installing the chart sn-platform in k8s v1.22+, the [PR](https://github.com/streamnative/charts/pull/447) had upgraded the deprecated apiVersion `extensions/v1beta1` of `Ingress` to `networking.k8s.io/v1`. But according to the [doc](https://kubernetes.io/blog/2021/07/26/update-with-ingress-nginx/), it will also need to upgrade the version of ingress nginx to `v1.x.x` for k8s v1.22. 

# Modifications
This PR is going to upgrade the ingress nginx image to use [v1.1.1](https://github.com/kubernetes/ingress-nginx/releases/tag/controller-v1.1.1) when the k8s version is v1.22+, and keep using `0.26.2` for the older versions.

According to the doc, only v1.1.1 supports k8s v1.23, so use this version directly.

But this pr will not use the new feature  `IngressClass` for now,  keep using the annotation `kubernetes.io/ingress.class: nginx`.

### DIFF
--- a/charts/sn-platform/templates/control-center/_control_center.tpl
+++ b/charts/sn-platform/templates/control-center/_control_center.tpl
@@ -85,6 +85,15 @@ pulsar controller ingress target port for http endpoint
     {{- default .Capabilities.KubeVersion.Version .Values.kubeVersionOverride -}}
 {{- end -}}
 
+{{/* Check Kubernetes version is less than v1.22 */}}
+{{- define "pulsar.kubeVersion.isLessThanV122" -}}
+    {{- if semverCompare "< 1.22-0" (include "pulsar.kubeVersion" .) -}}
+        {{- print "true" -}}
+    {{- else -}}
+        {{- print "false" -}}
+    {{- end -}}
+{{- end -}}
+
 
 {{/* Check Ingress API version is stable or not */}}
 {{- define "pulsar.ingress.isStable" -}}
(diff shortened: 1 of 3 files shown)


charts/sn-platform/templates/control-center/_control_center.tpl
```diff
@@ -93,4 +102,13 @@ pulsar controller ingress target port for http endpoint
...
         {{- print "false" -}}
     {{- end -}}
-{{- end -}}
\ No newline at end of file
+{{- end -}}
+
+
+{{- define "pulsar.ingress.image" -}}
+    {{- if and (eq (include "pulsar.kubeVersion.isLessThanV122" .) "false") (semverCompare "< 1.0.0" .Values.images.nginx_ingress_controller.tag )}}
+        {{- print "k8s.gcr.io/ingress-nginx/controller:v1.1.1"}}
```
fantapsody, line 110: It's a bit hacky, could you add some comments here?

fantapsody (approved):

charts/sn-platform/templates/control-center/_control_center.tpl
FushuWang: Done

tuteng (approved):

PR #469, opened 2022-02-05
[sn-platform]Support to add customized nodeAffinity and podAntiAffinity for JD  scenarios

FushuWang:
The integration of JD JDos, required to use specific topology key for pod zonal or regional availability.
Resolve https://github.com/streamnative/jd-cloud/issues/1


charts/sn-platform/templates/bookkeeper/bookkeeper-cluster.yaml
```diff
@@ -66,7 +66,10 @@ spec:
       {{- if and .Values.affinity.anti_affinity .Values.bookkeeper.affinity.anti_affinity }}
       podAntiAffinity:
         {{ .Values.bookkeeper.affinity.type }}:
-          {{ if eq .Values.bookkeeper.affinity.type "requiredDuringSchedulingIgnoredDuringExecution" }}
+          {{- if eq .Values.bookkeeper.affinity.type "requiredDuringSchedulingIgnoredDuringExecution" }}
+          {{- with .Values.bookkeeper.affinity.podAffinityTerms }}
```
fantapsody, line 70: Would it be better to move these custom affinity terms out of the `if` block? @ericsyh

charts/sn-platform/values.yaml
```diff
@@ -600,6 +600,40 @@ zookeeper:
     # requiredDuringSchedulingIgnoredDuringExecution - rules must be met for pod to be scheduled (hard) requires at least one node per replica
     # preferredDuringSchedulingIgnoredDuringExecution - scheduler will try to enforce but not guranentee
     type: preferredDuringSchedulingIgnoredDuringExecution
+
```
ericsyh, line 603: @FushuWang @fantapsody Can we just provide a custom affinity roles config like 

```
affinity:
    anti_affinity: true
    type: preferredDuringSchedulingIgnoredDuringExecution

    customAffinity: {}
```

to override the default rule of `spec.affinity`? No matter users want to set `podAffinity`,`nodeAffinity` or even mixed affinity rules.

ericsyh: For example like 

```
affinity:
    anti_affinity: true
    type: preferredDuringSchedulingIgnoredDuringExecution

    customAffinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: kubernetes.io/os
                operator: In
                values:
                - linux
          preferredDuringSchedulingIgnoredDuringExecution:
          - weight: 1
            preference:
              matchExpressions:
              - key: another-node-label-key
                operator: In
                values:
                - another-node-label-value
        podAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: security
                operator: In
                values:
                - S1
            topologyKey: topology.kubernetes.io/zone
```

fantapsody: @ericsyh Yeah, I think we should allow users to set these rules freely as needed.

FushuWang:
Close PR. Because there is a new approach. Use the `affinity.customRules` to do it.

PR #484, opened 2022-02-24
Fix Datadog annotation for Autodiscovery.

MarvinCai:
Update container name to the name used by operator if operator is enabled.

Resolve: https://github.com/streamnative/charts/issues/483


sijie (commented):
@MarvinCai do we need to change the function worker?

MarvinCai:
dedicated function worker is not managed by operator, also currently there's no datadog config for function worker
if we want to add that we can probably have another pr for that.

sijie (approved):

sijie:
Add `doc-required` label as we might need to document how sn-platform can integrate datadog for metrics autodiscovery.

Huanli-Meng:
The doc is updated. Here is the [PR](https://github.com/streamnative/doc-platform/pull/381). Therefore, remove the doc-required label and add doc-added label.

PR #487, opened 2022-02-24
Support no vault for auth

tuteng:


charts/sn-platform/values.yaml
```diff
@@ -197,7 +197,7 @@ images:
     pullPolicy: IfNotPresent
   streamnative_console:
     repository: streamnative/sn-platform-console
-    tag: "1.9"
+    tag: "1.10-rc13"
```
tuteng, line 200: 1.10 should be used instead of it

charts/sn-platform/templates/streamnative-console/streamnative-console-initialize.yaml
```diff
@@ -60,9 +60,17 @@ spec:
         {{- end }}
         - name: STREAMNATIVE_CONSOLE_BACKEND_URL
           value: http://{{ template "pulsar.streamnative_console.service" . }}:7750
+        {{- if not .Values.auth.vault.enabled }}
+        - name: SUPER_USER_NAME
+          value: admin
+        - name: SUPER_USER_PASSWORD
```
tuteng, line 66: When vault auth is not enabled, we can initialize the environment with an arbitrary account and password

tuteng:
the console needs to be upgraded to 1.10, I think this feature https://github.com/streamnative/charts/pull/487 needs more testing and we move it to 1.2.35

MarvinCai (approved):

PR #499, opened 2022-03-01
Fix datadog annotation for bk/zk/broker/proxy

MarvinCai:
### Motivation
Fix datadog annotation for bk/zk/broker/proxy

### Modifications
Datadog annotation annotation key is missing 

### Verifying this change

### Documentation

Check the box below.

Need to update docs? 
- [x] no-need-doc 
It's a bug fix.


sijie (approved):

sijie:
@MarvinCai can you add an integration test to verify this change? i.e you can verify if `helm template` works or not. This should be sufficient to catch any syntax errorrs.

PR #509, opened 2022-03-01
[charts/pulsar] Fix issues with openshift deployments

### LINKED ISSUE #507: [charts/pulsar] issue with the control center ingress file when deploying on openshift
### Issue
___
1) On line 25 of this file [control-center-ingress.yaml](https://github.com/streamnative/charts/blob/master/charts/pulsar/templates/control-center/control-center-ingress.yaml#L25) it has a hardcoded `apiversion` that according to openshift is deprecated (I have added the screenshot of the deprecated api info) and is causing issues.

2) On line 47 of this file [control-center-ingress.yaml](https://github.com/streamnative/charts/blob/master/charts/pulsar/templates/control-center/control-center-ingress.yaml#L47) it has a hardcoded annotation of `kubernetes.io/ingress.class: nginx`. This is causing issues when deploying on openshift. By having this in there openshift isn't automagically creating the route based on the information from the ingress object. 

3) On line 4 of the file [control_center.tpl](https://github.com/streamnative/charts/blob/master/charts/pulsar/templates/control-center/_control_center.tpl#L4) it defines the FQDN on the host that is to be used for the ingress object. This format is not correct for supporting openshift. Openshift looks for this format `<name>-<namespace>.apps.<your.cluster.domain>`  for the host name to be used. 

### Solution
___
 
1) Like in [charts/sn-platform] file on lines 22 and 28 - 32 in file [control-center-ingress.yaml](https://github.com/streamnative/charts/blob/master/charts/sn-platform/templates/control-center/control-center-ingress.yaml) provide some method like an `if` statement for toggling which `apiversion` to use based on what is available.

2) With an `if` statement identifying if it is openshift or not and allowing that to be toggled between true and false. If it is true then do not include the hardcoded annotation of `kubernetes.io/ingress.class: nginx`.

3) Again with an `if` statement identifying if it is openshift or not and allowing that to be toggled between true and false. If it is true then set the FQDN to this format `<name>-<namespace>.apps.<your.cluster.domain>`.

bsheltonihs:
Fixes #507 

### Motivation

Addressing issues with control center's the api version, annotations, and the FQDN that gets built so that this can be deployed on openshift.   

### Modifications

1) Added logic around the `apiversion` in the file `control-center-ingress.yaml` by using the value of `deployment.openshift` being true or false.

2) Added the same logic around the annotation `kubernetes.io/ingress.class: nginx` to address how openshift handles ingress objects and allowing for routes to be created.

3) Added logic in file `_control_center_tpl`  in section `define pulsar.control_center_domain` to build the URL FQDN that openshift expects which also builds the routes.

### Verifying this change

### Documentation

Check the box below.

Need to update docs? 
  
- [x] `no-need-doc` 
  
  I made comments in the value.yaml for what this new value does. At this point it is only a few items to change for the openshift deployment. If this were to grow I could see needing more information and documentation.

### DIFF
--- a/charts/pulsar/templates/control-center/_control_center.tpl
+++ b/charts/pulsar/templates/control-center/_control_center.tpl
@@ -3,6 +3,9 @@ control center domain
 */}}
 {{- define "pulsar.control_center_domain" -}}
 {{- if .Values.ingress.control_center.enabled -}}
+    {{- if .Values.deployment.openshift -}}
+{{- printf "%s-%s.%s" .Values.pulsar.pulsar_manager.service .Values.pulsar.namespace .Values.domain.suffix -}}
+    {{- else -}}
     {{- if .Values.ingress.control_center.external_domain }}
 {{- printf "%s" .Values.ingress.control_center.external_domain -}}
     {{- else -}}
--- a/charts/pulsar/templates/control-center/_control_center.tpl
+++ b/charts/pulsar/templates/control-center/_control_center.tpl
@@ -3,13 +3,14 @@ control center domain
 */}}
 {{- define "pulsar.control_center_domain" -}}
 {{- if .Values.ingress.control_center.enabled -}}
-    {{- if .Values.deployment.openshift -}}
-{{- printf "%s-%s.%s" .Values.pulsar.pulsar_manager.service .Values.pulsar.namespace .Values.domain.suffix -}}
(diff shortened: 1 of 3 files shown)


MarvinCai (approved):
lgtm

MarvinCai:
thanks for your contribution

PR #519, opened 2022-03-07
 Initialize pulsar common module

tuteng:
* Initialize pulsar common module

### TBD
- [ ] Add document
- [ ] Add test


tuteng:
This fixed https://github.com/streamnative/charts/pull/730

PR #521, opened 2022-03-09
[charts/pulsar] Add ability to specify the cert-manager apiVersion

pavel-hladik:
### Motivation

Without enabled TLS the helm-controller refuses the helmrelease by the reason:
```
      Helm install failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:
        line 43: mapping key "brokerServiceURL" already defined at line 41
        line 44: mapping key "brokerWebServiceURL" already defined at line 42
```
This should be solved to enable:
```
    tls:
      enabled: true
      broker:
        enabled: true
```
The condition is [here](https://github.com/streamnative/charts/blob/master/charts/pulsar/templates/proxy/proxy-configmap.yaml#L58-L61). But when trying to use the cert-manager and internal_issuer, the certs are not issued because the templates are using deprecated apiVersion `cert-manager.io/v1alpha2`.

### Modifications

Adding ability to specify the cert-manager apiVersion.

### Documentation

- [x] no-need-doc

I'm not sure if this change needs a docs change.

### DIFF
--- a/charts/pulsar/templates/tls/tls-cert-internal-issuer.yaml
+++ b/charts/pulsar/templates/tls/tls-cert-internal-issuer.yaml
@@ -19,7 +19,7 @@
 
 {{- if .Values.certs.internal_issuer.enabled }}
 {{- if eq .Values.certs.internal_issuer.type "selfsigning" }}
-apiVersion: cert-manager.io/v1alpha2
+apiVersion: "{{ .Values.certs.api_version }}"
 kind: Issuer
 metadata:
   name: "{{ template "pulsar.fullname" . }}-{{ .Values.certs.internal_issuer.component }}"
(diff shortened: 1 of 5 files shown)


bsheltonihs:
The one question I have is if the schema for the spec section is correct in the "tls-certs-internal.yaml" if you use the newer cert manager?

pavel-hladik:
> The one question I have is if the schema for the spec section is correct in the "tls-certs-internal.yaml" if you use the newer cert manager?
> 
> From what I have found these items in the spec section are not valid with the newer cert-manager.io/v1:
> 
> * `organization` ( it seems to be under `subject.organizations` )
> * `keySize` ( it seems to be under `privateKey.size` )
> * `keyAlgorithm` ( it seems to be under `privateKey.algorithm` )
> * `keyEncoding` ( it seems to be under `privateKey.encoding` )

Hi, pls see my update.

Huanli-Meng:
@sijie could you please help verify whether we need to add/update doc for this PR? Thanks

hadican:
Hi! Any update on this?

maxsxu:
Close this in favor of #807

PR #526, opened 2022-03-11
Expose console backend port

### LINKED ISSUE #522: sn-platform chart streamnative-console-init job stuck in endless loop
When deploying pulsar with the sn-platform chart, the streamnative-console-init job never completes even though the pulsar cluster deployment is up and running.
This is causing an issue when deploying pulsar with FLuxCD as its reconciliation loop considers the deployment to be incomplete, causing it to uninstall and redeploy pulsar in a loop, until the max retry is reached.

FluxCD complains about 1 job that is not finished to rule the deployment not ready:

```
Job is not completed: pulsar/pulsar-sn-platform-streamnative-console-init
warning: Upgrade "pulsar" failed: timed out waiting for the condition
  Normal  error  36s (x87 over 30h)  helm-controller  reconciliation failed: upgrade retries exhausted
```

tuteng:
Fixed #522 

Master Issue: #<xyz>

### Motivation

### Modifications

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/charts/sn-platform/conf/streamnative_console/init_instance.py
+++ b/charts/sn-platform/conf/streamnative_console/init_instance.py
@@ -9,8 +9,8 @@ if not username or not password:
     username = os.getenv("SUPER_USER_NAME")
     password = os.getenv("SUPER_USER_PASSWORD")
 loginData = {
-    "username": os.getenv("VAULT_SUPER_USER_NAME"),
-    "password": os.getenv("VAULT_SUPER_USER_PASSWORD"),
+    "username": username,
+    "password": password,
     "type":"pulsar-vault-userpass"
 }
 headers = {
--- a/charts/sn-platform/conf/streamnative_console/init_instance.py
+++ b/charts/sn-platform/conf/streamnative_console/init_instance.py
@@ -3,6 +3,11 @@ import json
 import os
 host = os.getenv("STREAMNATIVE_CONSOLE_BACKEND_URL")
 loginPath = "/cloud-manager/login"
+username = os.getenv("VAULT_SUPER_USER_NAME")
+password = os.getenv("VAULT_SUPER_USER_PASSWORD")
+if not username or not password:
+    username = os.getenv("SUPER_USER_NAME")
+    password = os.getenv("SUPER_USER_PASSWORD")
 loginData = {
     "username": os.getenv("VAULT_SUPER_USER_NAME"),
     "password": os.getenv("VAULT_SUPER_USER_PASSWORD"),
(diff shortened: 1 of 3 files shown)


charts/sn-platform/conf/streamnative_console/init_instance.py
```diff
@@ -3,6 +3,11 @@
 import os
 host = os.getenv("STREAMNATIVE_CONSOLE_BACKEND_URL")
 loginPath = "/cloud-manager/login"
+username = os.getenv("VAULT_SUPER_USER_NAME")
```
fantapsody, line 6: Where are these two variables used?

charts/sn-platform/conf/streamnative_console/init_instance.py
```diff
@@ -19,11 +24,17 @@
...
-    instanceData["service"] = json.dumps({"kopServiceUrl": os.getenv("KOP_SERVICE_URL")})
+    serviceData["kopServiceUrl"] = os.getenv("KOP_SERVICE_URL")
+if (os.getenv("FLINK_SERVICE_URL")):
+    serviceData["flinkServiceUrl"] = os.getenv("FLINK_SERVICE_URL")
+instanceData["service"] = json.dumps(serviceData)
+
 instancePath = "/cloud-manager/v1/instances"
 instanceResponse = requests.post(url = host + instancePath, data = json.dumps(instanceData), headers=headers)
-if (instanceResponse.ok):
+if (instanceResponse.ok and 'Add environment success' in instanceResponse.text):
```
fantapsody, line 36: It might not be ideal to rely on the message to check whether an API call succeeds. Could you give an example that `instanceResponse.ok` is true but adding environment failed?

charts/sn-platform/templates/streamnative-console/streamnative-console-service.yaml
```diff
@@ -41,7 +41,11 @@ spec:
     - name: frontend
       port: {{ .Values.streamnative_console.service.ports.frontend }}
       protocol: TCP
-      targetPort: {{ .Values.streamnative_console.ports.frontend }}
+      targetPort: 9527
```
fantapsody, line 44: Could `targetPort` use predefined values as well?

charts/sn-platform/conf/streamnative_console/init_instance.py
tuteng: Fixed

charts/sn-platform/conf/streamnative_console/init_instance.py
tuteng: A health check controller will be added later

charts/sn-platform/templates/streamnative-console/streamnative-console-service.yaml
tuteng: I currently recommend not using this predefined value, because currently the backend service starts on port 7750 and the frontend service starts on port 9527, the backend service port is configured in the configuration file and the frontend port is configured in nginx, if we change it, we need to do some work, if we configure it with the predefined value but the port is not really used in the service, it may cause the request not to be forwarded to the service, so I'll create an issue, and we'll implement it later

fantapsody (approved):

sijie:
@tuteng which release includes this fix?

tuteng:
@sijie I think it is sn-platform:1.2.38

PR #528, opened 2022-03-14
[charts/sn-platform] Support exposing broker and kop on istio

fantapsody:
Fixes https://github.com/streamnative/charts/issues/488

### Motivation

This PR supports exposing broker and kop services using istio.

### Modifications

New feature flags and configurations:
- `istio.enabled` as a flag to enable/disable istio support
- `istio.labels` for the istio labels to inject on pods.
- `istio.gateway.selector` for the selector of the gateway
- `istio.gateway.tls.certSecretName` for the secret of the certificate used for the gateway
- `ingress.broker.external_domain` for the cluster service URL exposed for broker and kop

Istio related changes
- Fix istio configs on `PulsarBroker`
- Add labels to inject istio sidecars
- Disable init containers that connect to other pods when istio is on to support mTLS
- Add authorization policies to allow network requests between pods when istio is on
- Add commands (`curl` or `nc` depends on the availability of corresponding tools in the image) to stop the istio proxy container automatically for jobs
- Expose control center through istio gateway when istio is on

Other changes
- Fix the service account usage in the bookkeeper cluster
- Allow using the specific service account in the zookeeper cluster
- Turn on `Autorecovery` by default
- Upgrade sn-platform images to `2.9.2.x` for multi-listener support
- Upgrade kafka container of toolset to `7.0.1`

TODOs
- integration with monitoring systems has not been tested 

### Verifying this change

### Documentation

Check the box below.

Need to update docs? 

- [x] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### Example values to enable istio
```
components:
  proxy: false
  kop: true

monitoring:
#  # monitoring - prometheus
  prometheus: false
#  # monitoring - grafana
  grafana: false
#  # monitoring - node_exporter
  node_exporter: false
#  # alerting - alert-manager
  alert_manager: false

auth:
  authentication:
    enabled: true
  authorization:
    enabled: true

kop:
  auth:
    enabled: true

initialize: true

domain:
  enabled: true

ingress:
  enabled: false
  control_center:
    enabled: true
    external_domain: admin.sn.io
  broker:
    enabled: true
    external_domain: test.sn.io

istio:
  enabled: true
  labels:
    istio.io/rev: stable
  gateway:
    selector:
      role: istio-ingressgateway
    tls:
      certSecretName: istio-tls
```


charts/sn-platform/templates/zookeeper/zookeeper-authorizationpolicy.yaml
```diff
@@ -0,0 +1,41 @@
...
+kind: AuthorizationPolicy
+metadata:
+  name: "{{ template "pulsar.fullname" . }}-{{ .Values.zookeeper.component }}"
+  namespace: {{ template "pulsar.namespace" . }}
+spec:
+  rules:
+  - to:
+    - operation:
+        ports:
+        - "2181"
```
MarvinCai, line 31: even if these ports are kind of stable and fixed, I think it's still good to read from value file so when we want to update then we can simply update the in value file rather then update the chart version.

charts/sn-platform/templates/broker/broker-authorizationpolicy.yaml
```diff
@@ -0,0 +1,42 @@
...
+metadata:
+  name: "{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}"
+  namespace: {{ template "pulsar.namespace" . }}
+spec:
+  rules:
+  - to:
+    - operation:
+        ports:
+        - "6650"
+        - "6652"
```
tuteng, line 32: Is 6652 the port for the tls of the pulsar service?

charts/sn-platform/templates/vault/vault-initialize.yaml
```diff
@@ -61,7 +64,8 @@ spec:
            done;
            echo "vault is ready~";
 
-           cd /root/pulsar/create_pulsar_tokens && /usr/local/bin/zsh create_pulsar_tokens.sh -n {{ template "pulsar.namespace" . }} -k {{ .Release.Name }} -v {{ template "pulsar.vault.url" . }} -r $ROOT_TOKEN && cd /root/pulsar/init_vault_streamnative_console && /usr/local/bin/zsh startup.sh
+           cd /root/pulsar/create_pulsar_tokens && /usr/local/bin/zsh create_pulsar_tokens.sh -n {{ template "pulsar.namespace" . }} -k {{ .Release.Name }} -v {{ template "pulsar.vault.url" . }} -r $ROOT_TOKEN && cd /root/pulsar/init_vault_streamnative_console && /usr/local/bin/zsh startup.sh;
+           echo -ne "POST /quitquitquit HTTP/1.1\r\nHost: 127.0.0.1:15020\r\nUser-Agent: curl/7.68.0\r\nAccept: */*\r\n\r\n"|nc 127.0.0.1:15020 || true;
```
MarvinCai, line 68: is the port fixed?

MarvinCai (approved):
overall LGTM

charts/sn-platform/values.yaml
```diff
@@ -79,7 +79,7 @@ components:
   # bookkeeper
   bookkeeper: true
   # bookkeeper - autorecovery
-  autorecovery: false
+  autorecovery: true
```
tuteng, line 82: Does this configuration need to be enabled by default?

charts/sn-platform/templates/broker/broker-authorizationpolicy.yaml
fantapsody: I would say yes because port 6652 serves requests from the gateway. While on the other hand, the gateway terminates TLS, and pulsar itself just serves plaintext requests.

charts/sn-platform/templates/streamnative-console/streamnative-console-authorizationpolicy.yaml
```diff
@@ -0,0 +1,39 @@
...
+  rules:
+  - to:
+    - operation:
+        ports:
+        - "7750"
+        - "9527"
+  action: ALLOW
+  selector:
+    matchLabels:
+      app: {{ template "pulsar.name" . }}
```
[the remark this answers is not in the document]
fantapsody: Pods created by pulsar operators use the label `cloud.streamnative.io/app`, while other pods are just using `app`. Maybe we should add the namespace to all labels we defined, but it's difficult to update the selectors of existing resources.

charts/sn-platform/templates/zookeeper/zookeeper-authorizationpolicy.yaml
fantapsody: The update to this port list may be more like a bug fix as they should be corresponding to ports exposed on pods that cannot be updated so far. I suggest we make them configurable only when we are sure it should be to keep it simple.

charts/sn-platform/templates/vault/vault-initialize.yaml
fantapsody: It seems to be: https://github.com/istio/istio/issues/6324#issuecomment-760156652

charts/sn-platform/values.yaml
fantapsody: I wonder why it wasn't enabled by default as it is used in ledger recovery and scaling down of the cluster, and I just turned it on. @ericsyh What's your opinion?

charts/sn-platform/values.yaml
```diff
@@ -2368,3 +2384,13 @@ entities:
...
+
+istio:
+  enabled: false
+  # istio labels used to inject sidecars if it's not `sidecar.istio.io/inject: "true"`
+  labels: {}
+  gateway:
+    # gateway selector if it's not `istio: ingressgateway`
+    selector: {}
+    tls:
+      certSecretName: "[CERTIFICATE SECRET NAME IN ISTIO ROOT NAMESPACE"
```
tuteng, line 2396: A right bracket may be missing on the right side

tuteng (approved):

charts/sn-platform/values.yaml
ericsyh: I think it will be better to turn the autorecovery on by default cause some users (especially new to pulsar)may not understand the usage of autorecovery. It will increase the risk of data lose for these users when they scaling down the bk or unable to create new bk pod in some cases.

charts/sn-platform/templates/broker/_broker.tpl
```diff
@@ -380,3 +380,20 @@ Define gcs offload options mounts
 {{ .Values.broker.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+pulsar service domain
+*/}}
+{{- define "pulsar.broker_service_domain" -}}
+{{- if .Values.ingress.broker.enabled -}}
+  {{- if .Values.ingress.broker.external_domain }}
```
ericsyh, line 389: @fantapsody There is no `ingress.broker.external_domain` config on `values.yaml`

ericsyh:
Istio PR can fix https://github.com/streamnative/charts/issues/536

charts/sn-platform/values.yaml
```diff
@@ -158,36 +158,36 @@ images:
...
   function_worker:
     repository: streamnative/sn-platform
-    tag: "2.8.1.30"
+    tag: "2.9.2.5"
     pullPolicy: IfNotPresent
   # NOTE: allow overriding the toolset image
   toolset:
     repository: streamnative/sn-platform
-    tag: "2.8.1.30"
+    tag: "2.9.2.5"
```
tuteng, line 186: Can you update the image version to 2.9.2.9, which includes a fix for oidc token

tuteng (approved):

ericsyh:
@fantapsody List my review comments here:
* I think you need to remove the initContainer for PulsarProxy (like check-broker) when istio enabled
* Remove the configs of `broker.istio.gateway` in `values.yaml`
* Add the sample of `ingress.broker.external_domain` in `values.yaml`

fantapsody:
> @fantapsody List my review comments here:
> 
> * I think you need to remove the initContainer for PulsarProxy (like check-broker) when istio enabled
> * Remove the configs of `broker.istio.gateway` in `values.yaml`
> * Add the sample of `ingress.broker.external_domain` in `values.yaml`

@ericsyh I have fixed the config values, please take a look again. For the proxy, we may not support running the proxy on istio in 1.3.0 because of two points:
- the istio gateway could route requests instead of the proxy
- there is a bug in the proxy that it cannot forward HTTP requests correctly when mTLS is on

We can work on the support of proxy on istio once the bug is fixed and there are customer requests to do so.

ericsyh (approved):
LGTM

PR #561, opened 2022-03-28
Support the configmap name of builtin connectors with release name as a prefix

FushuWang:
### Motivation

In order to install multiple releases in the same namespace, add support for the config map name of builtin connectors with release name as a prefix

### Modifications

Add support for the config map name of builtin connectors with release name as a prefix

### Verifying this change

### Documentation

Check the box below.

Need to update docs? 
  
- [x] `no-need-doc` 
  
  (Please explain why)

### DIFF
--- a/charts/sn-platform/templates/broker/_broker.tpl
+++ b/charts/sn-platform/templates/broker/_broker.tpl
@@ -409,7 +409,7 @@ Default name of builtin connectors
 {{/*
 Get the name of builtin connector for function mesh
 */}}
-{{- define "pulsar.functionMeshBuiltinConnectors" -}}
+{{- define "pulsar.builtinConnectorsConfigMapName" -}}
 {{- if .Values.broker.functionmesh.builtinConnectors -}}
 {{- printf "%s" .Values.broker.functionmesh.builtinConnectors -}}
 {{- else -}}
--- a/charts/sn-platform/templates/broker/_broker.tpl
+++ b/charts/sn-platform/templates/broker/_broker.tpl
@@ -397,3 +397,22 @@ pulsar service domain
   {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Default name of builtin connectors
+*/}}
+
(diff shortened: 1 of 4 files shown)


MarvinCai (approved):

charts/sn-platform/templates/broker/_broker.tpl
```diff
@@ -396,4 +396,23 @@ pulsar service domain
...
+*/}}
+
+{{- define "pulsar.builtinConnectorName" -}}
+{{- print "builtin-connectors" -}}
+{{- end -}}
+
+{{/*
+Get the name of builtin connector for function mesh
+*/}}
+{{- define "pulsar.functionMeshBuiltinConnectors" -}}
```
fantapsody, line 412: Actually it's the name of the configmap for builtin connectors, would it be better for the name to be more explicit?

charts/sn-platform/templates/broker/broker-configmap-functionmesh.yaml
```diff
@@ -21,7 +21,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: "builtin-connectors"
+  name: {{ .Release.Name }}-{{ include "pulsar.builtinConnectorName" . }}
```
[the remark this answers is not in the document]
FushuWang: > Why not using `pulsar.functionMeshBuiltinConnectors` directly?

`pulsar.functionMeshBuiltinConnectors` is used for `builtinConnectorRef` which has two scenarios:
- It can be assigned a config map that is created by our chart as the default
- It can also be assigned with a customized config map that is created by the customer 
This configmap is created by our chart, so I used the a new template to generate the name

charts/sn-platform/templates/broker/_broker.tpl
FushuWang: Fixed with `pulsar.builtinConnectorsConfigMapName`

fantapsody (approved):

PR #594, opened 2022-04-06
Fixed secretRefs check

tuteng:
### Motivation

Fixed exception: 
```
client.go:98: [debug] creating 35 resource(s)
Error: PulsarBroker.pulsar.streamnative.io "pulsar-poc-sn-platform" is invalid: spec.pod.secretRefs: Invalid value: "null": spec.pod.secretRefs in body must be of type array: "null"
helm.go:75: [debug] PulsarBroker.pulsar.streamnative.io "pulsar-poc-sn-platform" is invalid: spec.pod.secretRefs: Invalid value: "null": spec.pod.secretRefs in body must be of type array: "null"
```

### Modifications

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/charts/sn-platform/templates/proxy/proxy-cluster.yaml
+++ b/charts/sn-platform/templates/proxy/proxy-cluster.yaml
@@ -57,8 +57,8 @@ spec:
 {{ toYaml . | indent 6 }}
 {{- end }}
 {{- end }}
+{{- if or .Values.proxy.readPublicKeyFromFile .Values.proxy.extraSecretRefs }}
     secretRefs:
-    {{- if .Values.proxy.readPublicKeyFromFile }}
     {{- if .Values.proxy.publicKeyPath }}
     - mountPath: {{ .Values.proxy.publicKeyPath }}
     {{- else }}
@@ -69,9 +69,9 @@ spec:
       {{- else }}
       secretName: {{ template "pulsar.fullname" . }}-{{ .Values.vault.component }}-public-key
       {{- end }}
-    {{- end }}
 {{- with .Values.proxy.extraSecretRefs }}
 {{ toYaml . | indent 4 }}
+{{- end }}
 {{- end }}
     affinity:
       {{- if .Values.proxy.affinity.customRules }}
--- a/charts/sn-platform/templates/broker/broker-cluster.yaml
+++ b/charts/sn-platform/templates/broker/broker-cluster.yaml
@@ -60,8 +60,8 @@ spec:
 {{ toYaml . | indent 6 }}
 {{- end }}
 {{- end }}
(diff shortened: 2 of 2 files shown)


charts/sn-platform/templates/proxy/proxy-cluster.yaml
```diff
@@ -57,8 +57,8 @@ spec:
 {{ toYaml . | indent 6 }}
 {{- end }}
 {{- end }}
+{{- if or .Values.proxy.readPublicKeyFromFile .Values.proxy.extraSecretRefs }}
     secretRefs:
-    {{- if .Values.proxy.readPublicKeyFromFile }}
     {{- if .Values.proxy.publicKeyPath }}
```
jdmaguire, line 62: I've experienced bugs like this before. The root cause of not noticing them is usually that the code isn't clear.

For example, here are five lines (if, template, else, template, end) that could be this one line:

```suggestion
    - mountPath: {{ default "/pulsar/vault/v1/identity/oidc/.well-known/keys" .Values.proxy.publicKeyPath }}
```

The lines below this and the broker-cluster.yaml (in two spots) could similarly be cleaned up.

Again, my motivation being that simpler templating leads to cleaner templating. Cleaner templating makes writing bug-free templates easier.

jdmaguire, line 62: I've experienced bugs like this before. The root cause of not noticing them is usually that the code isn't clear.

For example, here are five lines (if, template, else, template, end) that could be this one line:

```
- mountPath: {{ default "/pulsar/vault/v1/identity/oidc/.well-known/keys" .Values.proxy.publicKeyPath }}
```

The lines below this and the broker-cluster.yaml (in two spots) could similarly be cleaned up.

Again, my motivation being that simpler templating leads to cleaner templating. Cleaner templating makes writing bug-free templates easier.

jdmaguire: I do like your fix and would approve it as-is but I think we should always take the opportunity to cleanup code when it arrives and is minor.

tuteng: Thanks, look forward to your changes!

Huanli-Meng:
@guangning, could you please help to verify whether we need to add/update docs for this PR? Thanks.

tuteng:
> @Guangning, could you please help to verify whether we need to add/update docs for this PR? Thanks.

I don't think this change needs documentation

jdmaguire (approved):

PR #602, opened 2022-04-07
[charts/sn-platform] Example Simplify Of Charts

jdmaguire:
### Motivation

I've been working with Helm for around five years now. I've seen lots of mistakes, made lots of mistakes, seen a lot of fixes for templates, and made a number myself.

I've discovered in this that the easiest way to avoid template bugs is to have simpler templates. This is particularly because Helm isn't expressive and YAML cares deeply for whitespace.

[Yesterday I saw a PR](https://github.com/streamnative/charts/pull/594). The PR was great. The bug it was fixing was sadly a bug I see somewhat often. An `if` condition in the wrong place. I see this happen often in long Helm templates, especially with nested blocks. A way to avoid this pitfall is to not have as many nests.

John Tip \#1 for Helm Templating: 

Prefer this:

```
field: {{ default "foo" .Values.bar }}
```

over this:

```
{{ if .Values.bar }}
field: {{ .Values.bar }}
{{ else }}
field: "foo"
{{ end }}
```

I wanted to show how this looks in a small section of code. I suggest against going through the Helm charts to clean them up but I think we can slowly simplify (and therefore) improve the Helm charts by doing simplifications like this.

It's a lot more easier write eights correct lines of code with one if than it is to write twenty correct lines with three ifs.

I will simplify you

### Modifications

I've reduced the code by using the 'default' function in Sprig. This replaces five lines of `if template else template end` with one line of `default x y` typically.

### Verifying this change

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/charts/sn-platform/templates/broker/broker-cluster.yaml
+++ b/charts/sn-platform/templates/broker/broker-cluster.yaml
@@ -62,16 +62,9 @@ spec:
 {{- end }}
 {{- if or .Values.broker.readPublicKeyFromFile (and .Values.broker.offload.gcs.enabled .Values.broker.offload.gcs.secret) .Values.broker.extraSecretRefs }}
     secretRefs:
-    {{- if .Values.broker.publicKeyPath }}
-    - mountPath: {{ .Values.broker.publicKeyPath }}
-    {{- else }}
-    - mountPath: /pulsar/vault/v1/identity/oidc/.well-known/keys
-    {{- end }}
-      {{- if .Values.broker.publicKeySecret }}
-      secretName: {{ .Values.broker.publicKeySecret }}
-      {{- else }}
-      secretName: {{ template "pulsar.fullname" . }}-{{ .Values.vault.component }}-public-key
-      {{- end }}
+    - mountPath: {{ default "/pulsar/vault/v1/identity/oidc/.well-known/keys" .Values.proxy.publicKeyPath }}
+      {{ $defaultSecretName := print (include "pulsar.fullname" .) "-" .Values.vault.component "-public-key" }}
+      secretName: {{ default $defaultSecretName .Values.proxy.publicKeySecret }}
     {{- if .Values.broker.offload.gcs.secret }}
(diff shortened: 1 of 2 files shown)


jdmaguire:
I could be entirely out to lunch ;) We all have our coding styles. Feel free to reject / close this PR.

My main intent is that if some of y'all like this, I'd like to propose a style guide for our charts later. The charts are got but I think they could be more readable and more maintainable.

sijie:
@jdmaguire That is a great suggestion! @fantapsody we should adopt this convention.

charts/sn-platform/templates/broker/broker-cluster.yaml
```diff
@@ -62,20 +62,9 @@ spec:
...
-    - mountPath: {{ .Values.broker.publicKeyPath }}
-    {{- else }}
-    - mountPath: /pulsar/vault/v1/identity/oidc/.well-known/keys
-    {{- end }}
-      {{- if .Values.broker.publicKeySecret }}
-      secretName: {{ .Values.broker.publicKeySecret }}
-      {{- else }}
-      secretName: {{ template "pulsar.fullname" . }}-{{ .Values.vault.component }}-public-key
-      {{- end }}
-    {{- if .Values.broker.offload.gcs.secret }}
```
fantapsody, line 75: Is this block removed?

fantapsody (commented):
Thanks @jdmaguire , this looks really great! Just left one comment as it seems on code block is removed?

charts/sn-platform/templates/broker/broker-cluster.yaml
jdmaguire: Good catch, accidentally removed this.

maxsxu (approved):
LGTM. Awesome! Been meaning to do this myself!

fantapsody (approved):

PR #609, opened 2022-04-13
[SN-Platform] Support using custom cert via mounting custom cert secret.

### LINKED ISSUE #536: Support m-TLS in sn-platform using user supplied certificates
*Motivation*

Users would like to configure sn-platform to use mutual TLS authentication.

*Requirements*

- Users would like to use user-supplied certificates. They don't want to use the certificates generated by the cert-manager.
- They require the following changes:
   - [ ] Proxy is configured to use m-TLS.
   - [ ] Broker is configured to use m-TLS. Proxy talks to Broker using m-TLS.
   - [ ] Bookie is configured to use m-TLS. Broker talks to Bookie using m-TLS.
   - [ ] ZooKeeper is configured to use m-TLS. Both brokers and bookies talk to zookeeper using m-TLS.
   - [ ] Function Worker is configured to use m-TLS. Function Worker talks to proxy using m-TLS.
   - [ ] Functions are configured to use m-TLS.

MarvinCai:
Fixes #536

### Motivation
User might have their own CA and sign certs by themself, we should support using user provided cert for components.

### Modifications
Add certSecretName for components which need to mount certs, so user can mount custom cert through secret.
Assuming there'll be only one CA and all cert secrets will be created in advance which will be out of scope of this chart.

### Verifying this change

### Documentation
- [x] `no-need-doc`


ericsyh:
@MarvinCai pls resove the conflicts.

charts/sn-platform/values.yaml
```diff
@@ -288,26 +290,33 @@ tls:
...
     cert_name: tls-zookeeper
+    certSecretName:
   # settings for generating certs for recovery
   autorecovery:
     cert_name: tls-recovery
+    certSecretName:
   # settings for generating certs for toolset
   toolset:
     cert_name: tls-toolset
+    certSecretName:
```
ericsyh, line 315: Shall we add some hints for users the difference between `cert_name` and `certSecretName` ?

ericsyh (approved):
LGTM

fantapsody (approved):

maxsxu (approved):

PR #616, opened 2022-04-18
Support custom config data for bookkeeper

tuteng:
Fixes #<xyz>

Master Issue: #<xyz>

### Motivation

* Some users complained that some custom configurations could not be loaded into the bookkeeper configuration  so this pr fixes it

### Modifications

* Support load custom configData for bookkeeper

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/charts/sn-platform/templates/bookkeeper/bookkeeper-cluster.yaml
+++ b/charts/sn-platform/templates/bookkeeper/bookkeeper-cluster.yaml
@@ -164,7 +164,11 @@ spec:
   {{- end }}
   config:
     custom:
+{{- if .Values.bookkeeper.custom }}
+{{ toYaml .Values.bookkeeper.custom | indent 6 }}
+{{- else }}
 {{ toYaml .Values.bookkeeper.configData | indent 6 }}
+{{- end }}
       zkLedgersRootPath: "{{ .Values.metadataPrefix }}/ledgers"
       # enable bookkeeper http server
       httpServerEnabled: "true"
--- a/charts/sn-platform/templates/bookkeeper/bookkeeper-cluster.yaml
+++ b/charts/sn-platform/templates/bookkeeper/bookkeeper-cluster.yaml
@@ -158,35 +158,31 @@ spec:
         {{- include "pulsar.bookkeeper.ledgers.storage.class" . | nindent 8 }}
     reclaimPolicy: Delete
   {{- end }}
-  {{- if .Values.bookkeeper.custom }}
   config:
     custom:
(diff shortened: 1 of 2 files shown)


charts/sn-platform/templates/bookkeeper/bookkeeper-cluster.yaml
```diff
@@ -158,34 +158,31 @@ spec:
         {{- include "pulsar.bookkeeper.ledgers.storage.class" . | nindent 8 }}
     reclaimPolicy: Delete
   {{- end }}
-  {{- if .Values.bookkeeper.custom }}
```
jdmaguire, line 161: With this gone, I see .Values.bookkeeper.custom no longer used.

If the intent is to put all custom config in configData, we could remove [.custom from the values file](https://github.com/streamnative/charts/blob/master/charts/sn-platform/values.yaml#L879).

tuteng:
@jdmaguire PTAL

Huanli-Meng:
@tuteng do we need to add/update doc for this PR?

Huanli-Meng:
> @tuteng do we need to add/update doc for this PR?

a `no-need-doc` check box is selected. Therefore, label this doc with `no-need-doc`

fantapsody (commented):
I'm not sure if any user is using `.Values.bookkeeper.custom`, removing it directly looks like a breaking change, or we can still put its value under `.spec.config.custom`? @ericsyh what's your opinion?

ericsyh:
> I'm not sure if any user is using `.Values.bookkeeper.custom`, removing it directly looks like a breaking change, or we can still put its value under `.spec.config.custom`? @ericsyh what's your opinion?

Yes, i think it will be better to keep compatibility in minor release and break on next major release.  Otherwise, using `.Values.bookkeeper.custom` on current SNP will get problem upgrade to 1.4

tuteng:
> > I'm not sure if any user is using `.Values.bookkeeper.custom`, removing it directly looks like a breaking change, or we can still put its value under `.spec.config.custom`? @ericsyh what's your opinion?
> 
> Yes, i think it will be better to keep compatibility in minor release and break on next major release. Otherwise, using `.Values.bookkeeper.custom` on current SNP will get problem upgrade to 1.4

Fixed @fantapsody @ericsyh PTAL

jdmaguire (approved):
Looks good and all comments addressed.

PR #619, opened 2022-04-19
[SN-Platform] Add sn-platform-stack.

### LINKED ISSUE #615: Create sn-platform-stack chart
As effort to modularizing sn-platform chart, we'll keep all core Pulsar components in sn-platform chart so they'll remain untouched during upgrade. So we'll need a new chart as parent chart for all sub-charts, thus we'll create a sn-platform-stack chart for this purpose.

We'll do it incrementally which means the  sn-platform-stack will just rely solely on sn-platform chart for now, and will be updated as the modularization work proceed.

MarvinCai:
Fixes #615

### Motivation
Create sn-platform-stack chart to prepare for sn-platform modularizing.

### Verifying this change

### Documentation
  
- [x] `no-need-doc`


charts/sn-platform-stack/values.yaml
```diff
@@ -0,0 +1,2336 @@
+#
```
jdmaguire, line 1: This is too big.

If sn-platform or sn-console has a reasonable default, you shouldn't duplicate the value here. The config here should only be the config that sn-platform-stack needs to change. It makes it harder to update the code and keep config in sync otherwise.

MarvinCai: make sense, will only keep necessary values here

fantapsody (approved):

PR #653, opened 2022-05-03
[charts/pulsar] adding flag webSocketServiceEnabled cm

yuweisung:
### Motivation

* According to the doc,  proxy supports websocket but the broker and proxy cm template have no option to turn it on.*

### Modifications

*The proxy-configmap.yaml and broker-configmap.yaml are modified to enable websocket services.*

### Verifying this change

### Documentation

Check the box below.

Need to update docs? 

  
- [ x ] `no-need-doc` 
  
  (Please explain why)

### DIFF
--- a/charts/pulsar/templates/broker/broker-configmap.yaml
+++ b/charts/pulsar/templates/broker/broker-configmap.yaml
@@ -42,7 +42,9 @@ data:
   numHttpServerThreads: "8"
   zooKeeperSessionTimeoutMillis: "30000"
   statusFilePath: "{{ template "pulsar.home" . }}/status"
-
+  {{- if .Values.proxy.websocket.enabled }}
+  webSocketServiceEnnabled: "true"
+  {{- end }}
   ## Offloading settings
   {{- if .Values.broker.offload.enabled }}
   offloadersDirectory: "{{ template "pulsar.home" . }}/offloaders"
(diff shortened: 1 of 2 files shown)


charts/pulsar/templates/broker/broker-configmap.yaml
```diff
@@ -42,7 +42,9 @@ data:
   numHttpServerThreads: "8"
   zooKeeperSessionTimeoutMillis: "30000"
   statusFilePath: "{{ template "pulsar.home" . }}/status"
-
+  {{- if .Values.proxy.websocket.enabled }}
+  webSocketServiceEnnabled: "true"
```
fantapsody, line 46:
```suggestion
  webSocketServiceEnabled: "true"
```

ericsyh:
@fantapsody I think there is no need to enable ws on broker, https://github.com/streamnative/charts/blob/master/charts/pulsar/values.yaml#L1291 allows to enable websocket on proxy.

polyzos:
@ericsyh what happens when we use istio and the proxy is disabled?

ericsyh:
> @ericsyh what happens when we use istio and the proxy is disabled?

@polyzos You mean the scenario that enabled Istio but wanna to use the websocket right? I think it's quite a good point, the proxy now has to been disabled cause there is a bug in the proxy that it cannot forward HTTP requests correctly when mTLS is on.  If this problem is fixed, proxy can work with Istio. 

So for the websocket, if currently no specific request for using websocket with Istio enabled, i prefer waiting for the proxy bug fixing.

PR #690, opened 2022-05-13
[sn-platform]Integrate cert-manager with Istio Ingress Gateway

### LINKED ISSUE #570: [feat][sn-platform]: integrate cert-manager with Istio gateway
Support cert-manager with Istio gateway:
* selfsigning
* public-cert-issuer
* Istio root namespace

ciiiii:
Resolves #570 

### Motivation

Generate Istio Ingress Gateway TLS cert with cert-manager

### Modifications
- Define Internal/Public Issuer and Cert for Istio Ingress Gateway TLS generate
  - Internal Issuer config: `.Values.certs.istio_internal_issuer`
  - Public Issuer config: `.Values.certs.istio_public_issuer`
- Use `.Values.istio.gateway.namespace` to specific Istio Ingress Gateway Namespace

**NOTE**: Since `PulsarBroker` generates Pod DNSName based on `advertisedDomain`(domain of Broker), there's a limit that `broker.external_domain` and `control_center.external_domain` must be covered by the `*.{{ .Values.domain.suffix }}` if they are defined.

### Verifying this change

### Documentation

Need to update docs? 

- [x] `doc-required`


charts/sn-platform/values.yaml
```diff
@@ -2353,4 +2364,9 @@ istio:
     # gateway selector if it's not `istio: ingressgateway`
     selector: {}
     tls:
+      # use existing secret
       certSecretName: "[CERTIFICATE SECRET NAME IN ISTIO ROOT NAMESPACE]"
+      # use cert-manager to generate secret, only take effect if certSecretName is empty,
+      # also need to config certs.istio_internal_issuer/certs.istio_public_issuer
+      cert_name: "[ISTIO CERT NAME]"
+      cert_namespace: "[ISTIO ROOT NAMESPACE]"
```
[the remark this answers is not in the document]
fantapsody: Should we leave the value empty otherwise the default value (`istio-system`) will not work?

charts/sn-platform/values.yaml
```diff
@@ -2353,4 +2364,9 @@ istio:
     # gateway selector if it's not `istio: ingressgateway`
     selector: {}
     tls:
+      # use existing secret
       certSecretName: "[CERTIFICATE SECRET NAME IN ISTIO ROOT NAMESPACE]"
+      # use cert-manager to generate secret, only take effect if certSecretName is empty,
+      # also need to config certs.istio_internal_issuer/certs.istio_public_issuer
+      cert_name: "[ISTIO CERT NAME]"
```
fantapsody, line 2371: If the customer needs to set a cert name anyway, why not just stick to the `certSecretName`?

charts/sn-platform/values.yaml
```diff
@@ -397,6 +397,17 @@ certs:
     enabled: false
     component: public-cert-issuer
     type: acme
+
+  istio_internal_issuer:
```
fantapsody, line 401: Is it possible to just reuse the existing `internal_issuer` and `public_issuer`?

charts/sn-platform/templates/tls/istio/tls-certs.yaml
```diff
@@ -0,0 +1,50 @@
...
+  usages:
+    - server auth
+    - client auth
+  dnsNames:
+  - "*.{{ .Values.domain.suffix }}"
+  issuerRef:
+{{- if .Values.certs.istio_internal_issuer.enabled }}  
+    name: "{{ template "pulsar.fullname" . }}-{{ .Values.certs.istio_internal_issuer.component }}"
+{{- end }}
+{{- if .Values.certs.istio_public_issuer.enabled }}
```
fantapsody, line 44: Would it be better to use something like an `else if` here?

charts/sn-platform/templates/tls/istio/tls-certs.yaml
```diff
@@ -0,0 +1,50 @@
...
+  isCA: false
+  privateKey:
+    algorithm: RSA
+    encoding: PKCS1
+    size: 2048
+  usages:
+    - server auth
+    - client auth
+  dnsNames:
+  - "*.{{ .Values.domain.suffix }}"
```
fantapsody, line 39: It covers only one case, while the user may also define `ingress.[control_center/broker].external_domain` as well.

charts/sn-platform/templates/tls/istio/tls-certs.yaml
```diff
@@ -0,0 +1,50 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+{{- if .Values.istio.enabled }}
+{{- if or .Values.certs.istio_internal_issuer.enabled .Values.certs.istio_public_issuer.enabled }}
```
fantapsody, line 20: If we can reuse the existing issuers, we may use another value as the flag to create this certificate.

charts/sn-platform/templates/tls/istio/tls-certs.yaml
ciiiii: Ref the previous public cert definition: https://github.com/streamnative/charts/blob/0bba32fc686a561e4280d36ed9ccaf698ca37fc1/charts/sn-platform/templates/tls/tls-certs-public.yaml#L32-L36
If `ingress.[control_center/broker].external_domain` is covered by `*.{{ .Values.domain.suffix }}`, Certificate will be invalid.

charts/sn-platform/values.yaml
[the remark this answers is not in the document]
ciiiii: > Seems camelcase is prefered for helm chart value names: https://helm.sh/docs/chart_best_practices/values/#naming-conventions

Follow eric's suggestion, I refer to the previous cert-related variables, should we fix all these values?
https://github.com/streamnative/charts/blob/91e4dcc9bbc1c50b5c91017e0ae569c209725d3b/charts/sn-platform/values.yaml#L283
https://github.com/streamnative/charts/blob/91e4dcc9bbc1c50b5c91017e0ae569c209725d3b/charts/sn-platform/values.yaml#L291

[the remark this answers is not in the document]
ciiiii: > Should we leave the value empty otherwise the default value (`istio-system`) will not work?

Yes, I will leave it empty.

charts/sn-platform/values.yaml
ciiiii: Maybe we don't need to unify Istio TLS configuration with others, how about the implementation in my first commit.

charts/sn-platform/values.yaml
ciiiii: If we change public_issuer and internal_issuer from `Issuer` to `ClusterIssuer`, it will be easy to reuse.

charts/sn-platform/values.yaml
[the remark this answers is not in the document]
fantapsody: > Follow eric's suggestion, I refer to the previous cert-related variables, should we fix all these values?

No need to rename existing names as it might cause compatibility issues, we just need to make it consistent for new names for now.

charts/sn-platform/values.yaml
fantapsody: Do you mean [this one](https://github.com/streamnative/charts/pull/690/commits/912f4d00bc9d3754f4eebeccc34636251f73b9d7#diff-f8fce96bac42e794714091b347ba9ba3152d535f10477277716a2172e2d8566bR6)? It looks mostly good to me.

charts/sn-platform/values.yaml
fantapsody: I see. We generally prefer namespace-scoped resources over cluster-scoped resources, so it's fine to create separate issuers here.

charts/sn-platform/templates/tls/istio/tls-certs.yaml
```diff
@@ -0,0 +1,59 @@
...
+    algorithm: RSA
+    encoding: PKCS1
+    size: 2048
+  usages:
+    - server auth
+    - client auth
+  dnsNames:
+  - "*.{{ .Values.domain.suffix }}"
+{{- if and .Values.components.broker .Values.ingress.broker.external_domain }}
+{{- if ne (hasSuffix (include "domain.suffix" .) .Values.ingress.broker.external_domain) true }}
```
fantapsody, line 41: why not use `not` here but `ne xxx true`?

fantapsody (approved):

PR #697, opened 2022-05-16
Support SecurityContextConstraints for deploying on OpenShift

### LINKED ISSUE #696: [sn-platform][feat] Enable SecurityContextConstraints for deploying on OpenShift
## Motivation
Similar to `PodSecurityPolicy` (`PSP` for short) in the native Kubernetes, OpenShift uses `SecurityContextConstraints` (`SCC` for short) to control permissions for pods. 

An OpenShift cluster contains several builtin SCCs, where the `restricted` SCC will be used by default for authenticated users. The following is a quote from the official OpenShift documentation：

> The `restricted` SCC denies access to all host features and requires pods to be run with a UID, and SELinux context that are allocated to the namespace. This is the most restrictive SCC provided by a new installation and will be used by default for authenticated users.
> The restricted SCC is the most restrictive of the SCCs that ship by default with the system.

So we need to create a custom SCC to allow the sn-platform to be successfully deployed on OpenShift.

## References
- https://docs.openshift.com/container-platform/4.10/authentication/managing-security-context-constraints.html

maxsxu:
Fixes #696 

### Motivation

To create a custom SCC to allow the sn-platform to be successfully deployed on OpenShift.

### Modifications

Added a new value `openshift.enabled`  to control generating SCC and related role/rolebinding.

### Verifying this change

```
openshift:
  enabled: true
```

### Documentation

Check the box below.

Need to update docs? 

- [x] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/charts/sn-platform/templates/openshift/_openshift.tpl
+++ b/charts/sn-platform/templates/openshift/_openshift.tpl
@@ -1,3 +1,3 @@
 {{- define "pulsar.openshift.scc.name" -}}
-{{ template "pulsar.fullname" . }}-openshift-scc
+{{ template "pulsar.namespace" . }}-{{ template "pulsar.fullname" . }}-openshift-scc
 {{- end -}}
--- a/charts/sn-platform/templates/openshift/_openshift.tpl
+++ b/charts/sn-platform/templates/openshift/_openshift.tpl
@@ -1,0 +1,3 @@
+{{- define "pulsar.openshift.scc.name" -}}
+{{ template "pulsar.fullname" . }}-openshift-scc
+{{- end -}}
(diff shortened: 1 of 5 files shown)


charts/sn-platform/templates/openshift/_openshift.tpl
```diff
@@ -0,0 +1,3 @@
+{{- define "pulsar.openshift.scc.name" -}}
+{{ template "pulsar.fullname" . }}-openshift-scc
```
fantapsody, line 2: Maybe it's better to add the namespace to the name to reduce potential conflicts, but it might be ok for now.

fantapsody (approved):

FushuWang (approved):

charts/sn-platform/templates/openshift/_openshift.tpl
maxsxu: Make sense to me. Updated

ericsyh:
@maxsxu Did this PR cherry-pick to 1.4?

Huanli-Meng:
The doc is added through this PR: https://github.com/streamnative/doc-platform/pull/451

PR #719, opened 2022-05-25
Support enable and disable connector flink sql on console

tuteng:

Fixes #<xyz>

Master Issue: #<xyz>

### Motivation

### Modifications

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

Check the box below.

Need to update docs? 

- [x] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/charts/sn-console/templates/console/streamnative-console-statefulset.yaml
+++ b/charts/sn-console/templates/console/streamnative-console-statefulset.yaml
@@ -165,9 +165,9 @@ spec:
             value: {{ .Values.configData.BACKEND_DEFAULT_SUPER_USER_ROLE }}
           {{- end }}
           - name: CONNECTOR_ENABLED
-            value: {{ .Values.configData.CONNECTOR_ENABLED | default "false" }}
+            value: "{{ .Values.configData.CONNECTOR_ENABLED | default "false" }}"
           - name: FLINK_SQL_ENABLED
-            value: {{ .Values.configData.FLINK_SQL_ENABLED | default "false" }}
+            value: "{{ .Values.configData.FLINK_SQL_ENABLED | default "false" }}"
           {{- if .Values.login.pulsarJwt.enabled }}
           - name: AUTHENTICATION_NAME
             value: "pulsar-jwt"
--- a/charts/sn-console/templates/console/streamnative-console-statefulset.yaml
+++ b/charts/sn-console/templates/console/streamnative-console-statefulset.yaml
@@ -165,7 +165,7 @@ spec:
             value: {{ .Values.configData.BACKEND_DEFAULT_SUPER_USER_ROLE }}
           {{- end }}
           - name: CONNECTOR_ENABLED
-            value: "{{ .Values.configData.CONNECTOR_ENABLED | default "false" }}"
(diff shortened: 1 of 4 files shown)


charts/sn-platform/values.yaml
```diff
@@ -1966,6 +1966,8 @@ streamnative_console:
     GOOGLE_REDIRECT_URI: ""
     GRAFANA_AUTH_PROXY: false
     GRAFANA_AUTH_PROXY_USER: ""
+    CONNECTOR_ENABLED: false
```
ericsyh, line 1969: we can set `CONNECTOR_ENABLED` true by default

charts/sn-platform/values.yaml
```diff
@@ -1966,6 +1966,8 @@ streamnative_console:
...
+    FLINK_SQL_ENABLED: false
```
ericsyh, line 1970: I don't think we should add `FLINK_SQL_ENABLED` on console now, cause we don't provide Flink Cluster on sn-platform now.

charts/sn-platform/values.yaml
tuteng: Fixed

ericsyh (approved):
LGTM

charts/sn-platform/values.yaml
tuteng: I removed the flink sql configuration `FLINK_SQL_ENABLED` from the value file, but kept the environment variables in statefulset, what do you think?

charts/sn-console/templates/console/streamnative-console-statefulset.yaml
```diff
@@ -164,6 +164,10 @@ spec:
           - name: BACKEND_DEFAULT_SUPER_USER_ROLE
             value: {{ .Values.configData.BACKEND_DEFAULT_SUPER_USER_ROLE }}
           {{- end }}
+          - name: CONNECTOR_ENABLED
+            value: "{{ .Values.configData.CONNECTOR_ENABLED | default "true" }}"
+          - name: FLINK_SQL_ENABLED
+            value: "{{ .Values.configData.FLINK_SQL_ENABLED | default "false" }}"
```
MarvinCai, line 170: shall we also add FLINK_SQL_ENABLED to value file?

tuteng: Yes, I think it's okay because we want some users to be able to support flink clusters very easily on the console after they've been deployed, but this feature is not directly exposed to users by default

maxsxu: Should we avoid this kind of ghost value? Thinking about whether we should have all `.Values` references correspond to the `values.yaml` file one by one?

MarvinCai (approved):

maxsxu (approved):

charts/sn-console/templates/console/streamnative-console-statefulset.yaml
tuteng: As mentioned above, it is possible to expose the FLINK_SQL_ENABLED option to the user is not suitable, because the sn-platform does not support creating a flink cluster.

PR #730, opened 2022-05-31
Create chart for Pulsar detector.

### LINKED ISSUE #460: Make chart for streamnative/sn-platform sla detector components.
Move sla detector components in streamnative/sn-platform to a new separate chart.
Need to provide proper name/config override to support adopting resources from existing helm release.

MarvinCai:
resolves #460

### Motivation
Create dedicated chart for Pulsar detector.

### Modifications
Create new chart sn-platform-detector to hold Pulsar detector component.
Make sn-platform-stack to use `sn-platform-detector` chart.

### Verifying this change

### Documentation

- [x] `no-need-doc`

### DIFF
--- a/charts/sn-platform-detector/templates/pulsar-detector-deployment.yaml
+++ b/charts/sn-platform-detector/templates/pulsar-detector-deployment.yaml
@@ -20,7 +20,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: "{{ template "pulsar.fullname" . }}-{{ .Values.pulsar_detector.component }}"
+  name: "{{ template "pulsar.fullname" . }}{{ template "pulsar.detector.component" . }}"
   namespace: {{ template "pulsar.namespace" . }}
   labels:
     {{- include "pulsar.standardLabels" . | nindent 4 }}
@@ -87,7 +87,7 @@ spec:
             done;
       {{- end }}
       containers:
-      - name: "{{ template "pulsar.fullname" . }}-{{ .Values.pulsar_detector.component }}"
+      - name: "{{ template "pulsar.fullname" . }}"
         image: "{{ .Values.images.pulsar_detector.repository }}:{{ .Values.images.pulsar_detector.tag }}"
         imagePullPolicy: {{ .Values.images.pulsar_detector.pullPolicy }}
         command: ["sh", "-c"]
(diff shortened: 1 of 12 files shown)


charts/sn-platform-detector/values.yaml
```diff
@@ -0,0 +1,81 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+## Namespace to deploy 
+namespace: ""
+stackName: pulsar-dector
```
FushuWang, line 22: It should be `global.stackName` according to the `_common.tpl`. En..., I remembered it was just `stackName` without `global` in the `_common.tpl` when reviewing the PR for`sn-platform-vault`.

charts/sn-platform-detector/templates/pulsar-detector-deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: "{{ template "pulsar.fullname" . }}-{{ .Values.pulsar_detector.component }}"
```
FushuWang, line 23: Do we still need the `.Values.pulsar_detector.component` as part of the name?  Now each chart has the component name as the suffix of the chart name, it would be repeated. I use the release name `ok` to render the manifests. The name is `ok-sn-platform-detector-pulsar-detector`.

charts/sn-platform-detector/values.yaml
MarvinCai: yeah, changed to `global.stackName` so in sn-platform-stack we only need to define once instead of passing stack name to each subchart.
actually don't need stackName when using independently, removing it.

charts/sn-platform-detector/templates/pulsar-detector-deployment.yaml
MarvinCai: that's true, let me see how we can improve

FushuWang (approved):

Issue #716, opened 2022-05-23
Support wal-compression option for sn-platform prometheus


tuteng:
Some users seem to have a lot of topics in the testing phase, which may generate a lot of monitoring data, and that may cause the prometheus service to restart due to the oom on wal replay, so we can consider adding the `--storage.tsdb.wal-compression` option to prometheus and making it enabled by default

fantapsody:
Extra arguments

PR #751, opened 2022-06-15
Default enable wal compression, for avoid prometheus oom

tuteng:
Master Issue: #716 

### Motivation

https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects

```
--storage.tsdb.wal-compression: Enables compression of the write-ahead log (WAL). Depending on your data, you can expect the WAL size to be halved with little extra cpu load. This flag was introduced in 2.11.0 and enabled by default in 2.20.0. Note that once enabled, downgrading Prometheus to a version below 2.11.0 will require deleting the WAL.
```

### Modifications

* Add option for enable wal compression for prometheus

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/charts/sn-platform/templates/prometheus/prometheus-statefulset.yaml
+++ b/charts/sn-platform/templates/prometheus/prometheus-statefulset.yaml
@@ -172,6 +172,9 @@ spec:
   {{ if .Values.prometheus.maxQuerySamples }}
           - --query.max-samples={{.Values.prometheus.maxQuerySamples }}
   {{- end }}
+  {{ if .Values.prometheus.args.walCompressionEnabled }}
+          - --storage.tsdb.wal-compression
+  {{- end }}
         ports:
         - name: server
           containerPort: {{ .Values.prometheus.port }}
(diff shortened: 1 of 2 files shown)


charts/sn-platform/templates/prometheus/prometheus-statefulset.yaml
```diff
@@ -171,6 +171,9 @@ spec:
           - --web.enable-lifecycle
   {{ if .Values.prometheus.maxQuerySamples }}
           - --query.max-samples={{.Values.prometheus.maxQuerySamples }}
+  {{- end }}
+  {{ if .Values.prometheus.args.walCompressionEnabled }}
+          - --storage.tsdb.wal-compression
   {{- end }}
```
ciiiii, line 177: Shall we also define a `.Values.prometheus.extraArgs` field? so we can custom args without updating chart next time.

tuteng: Thanks, I have fixed your comment, PTAL

ciiiii (approved):
LGTM

Huanli-Meng:
@tuteng Do we need to update the doc for this PR? Thanks.

tuteng:
> @tuteng Do we need to update the doc for this PR? Thanks.

Probably not right now

PR #790, opened 2022-07-04
[sn-platform] Improve function-worker compatibility

ciiiii:
### Motivation
- support Istio mode
- function worker compatible with TLS authentication

### Modifications
- add AuthorizationPolicy for function worker
- skip wait-proxy-ready initContainer when the proxy is disabled
- improve/fix function-worker TLS related configurations

### Verifying this change

### Documentation

Need to update docs? 
  
- [x] `no-need-doc`

### DIFF
--- a/charts/sn-platform/templates/broker/function-worker-configfile-configmap.yaml
+++ b/charts/sn-platform/templates/broker/function-worker-configfile-configmap.yaml
@@ -30,31 +30,32 @@ metadata:
 data:
   functions_worker.yml: |
     # Function package management
-    {{- if and .Values.tls.enabled .Values.tls.broker.enabled }}
+    workerPort: {{ .Values.functions.ports.http }}
+    {{- if and .Values.tls.enabled .Values.tls.functions.enabled }}
+    # TLS Settings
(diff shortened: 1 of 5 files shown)


charts/sn-platform/templates/broker/function-worker-configfile-configmap.yaml
```diff
@@ -30,32 +30,33 @@ metadata:
 data:
   functions_worker.yml: |
     # Function package management
-    {{- if and .Values.tls.enabled .Values.tls.broker.enabled }}
+    workerPort: {{ .Values.functions.ports.http }}
```
MarvinCai, line 33: is this compatible with non-istio based tls set up?

ciiiii: The TLS part configuration is not related to Istio, so it can work under non-istio env.

ciiiii: And previous worker template shouldn't work well with TLS mode; For example, `workerPort` is a required configuration in the current Pulsar version.

fantapsody (approved):

tuteng (approved):

PR #811, opened 2022-07-20
Update ZooKeeper dynamic reconfig.

MarvinCai:
### Motivation
Add skipACL if enabled dynamic reconfig for zookeeper.
We need to either skipACL or set a password, so use skipACL here.

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/charts/pulsar/templates/zookeeper/zookeeper-configmap.yaml
+++ b/charts/pulsar/templates/zookeeper/zookeeper-configmap.yaml
@@ -44,7 +44,6 @@ data:
   {{- end }}
   {{- if .Values.zookeeper.reconfig.enabled }}
   PULSAR_PREFIX_reconfigEnabled: "true"
-  PULSAR_PREFIX_quorumListenOnAllIPs: "true"
   PULSAR_PREFIX_skipACL: "yes"
   {{- end }}
   PULSAR_PREFIX_peerType: {{ .Values.zookeeper.peerType }}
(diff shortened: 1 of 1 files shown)


charts/pulsar/templates/zookeeper/zookeeper-configmap.yaml
```diff
@@ -45,6 +45,7 @@ data:
   {{- if .Values.zookeeper.reconfig.enabled }}
   PULSAR_PREFIX_reconfigEnabled: "true"
   PULSAR_PREFIX_quorumListenOnAllIPs: "true"
+  PULSAR_PREFIX_skipACL: "yes"
```
tuteng, line 48: Do you need to consider adding an option to control whether it is turned on or off, and is there a security risk in setting it to skipACL by default?

MarvinCai: if we need to enable dynamic reconfig then we need to either set skipACL or explicitly set a password, I think the risk is pretty much the same, and zk is not exposed to public network though

maxsxu (approved):

charts/pulsar/templates/zookeeper/zookeeper-configmap.yaml
```diff
@@ -45,6 +45,7 @@ data:
   {{- if .Values.zookeeper.reconfig.enabled }}
   PULSAR_PREFIX_reconfigEnabled: "true"
   PULSAR_PREFIX_quorumListenOnAllIPs: "true"
```
fantapsody, line 47: I heard this one might not be needed and may cause problems, could you do some research about this config item?

fantapsody (approved):

tuteng (approved):

PR #839, opened 2022-08-10
feat(sn-platform): Add labels and missing missing configs for Vault

ericsyh:
### Motivation

This PR made some enhancements to the Vault template

### Modifications

This PR changed Vault template including:
* Added `labels` config
* Added the missing config of pod level containerSecurityContext for `vault.banzaicloud.com/v1alpha1` crd
* Added the missing config of `annotations` for vault related Jobs
* Set env `SKIP_CHOWN`, `SKIP_SETCAP` as true and set config `disable_mlock` as true when Vault is running as non-root

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

charts/sn-platform/templates/vault/vault-clear-resource.yaml
```diff
@@ -37,6 +37,13 @@ spec:
     metadata:
       labels:
         {{- include "pulsar.template.labels" . | nindent 8 }}
+{{- if .Values.vault.labels }}    
+{{ toYaml .Values.vault.labels | indent 8 }}
+{{- end }}
+      annotations:
+{{- if .Values.vault.annotations }}
```
fantapsody, line 44: Will it be a problem if `.Values.vault.annotations` is nil?

charts/sn-platform/templates/vault/vault-statefulset.yaml
```diff
@@ -32,9 +32,20 @@ spec:
   serviceAccount: {{ template "pulsar.vault.serviceAccount" . }}
   serviceType: {{ .Values.vault.serviceType }}
   statsdDisabled: true
+  {{- if .Values.vault.securityContext }}
+  securityContext: {{- toYaml .Values.vault.securityContext | nindent 4 }}
+  vaultEnvsConfig:
```
fantapsody, line 37: Is this necessary?

charts/sn-platform/templates/vault/vault-statefulset.yaml
```diff
@@ -98,6 +114,9 @@ spec:
   # See https://www.vaultproject.io/docs/configuration/ for more information.
   config:
   {{ toYaml .Values.vault.config | nindent 4 }}
+  {{- if .Values.vault.securityContext }}
```
fantapsody, line 117: Is this necessary?

fantapsody (commented):
Is the motivation for the security context to support running as non-root or minimal permission mode?

charts/sn-platform/templates/vault/vault-clear-resource.yaml
ericsyh: You mean if the values config set like? 

```
vault:
  annotations: {}
```

charts/sn-platform/templates/vault/vault-statefulset.yaml
ericsyh: Without this config, seems the vault will be failed to start and print some `Operation not permitted` logs. 

I added these two Env configs from: https://github.com/hashicorp/docker-vault/issues/137#issuecomment-530247729

charts/sn-platform/templates/vault/vault-statefulset.yaml
ericsyh: Without this config, vault will fail to start and logs as 

```
Error initializing core: Failed to lock memory: cannot allocate memory

This usually means that the mlock syscall is not available.
Vault uses mlock to prevent memory from being swapped to
disk. This requires root privileges as well as a machine
that supports mlock. Please enable mlock on your system or
disable Vault from using it. To disable Vault from using it,
set the `disable_mlock` configuration option in your configuration
file.
2022-08-11T01:26:29.465Z [INFO]  proxy environment: http_proxy="" https_proxy="" no_proxy=""
```

But from the vault docs: https://www.vaultproject.io/docs/configuration#disable_mlock, seems we can use environment variable `VAULT_DISABLE_MLOCK` also.

ericsyh:
> Is the motivation for the security context to support running as non-root or minimal permission mode?

Oh yes. Seems some financial users have strict admission policies for running as non-root or minimal permission mode. I think OpenShift also has similar requirements.

fantapsody:
> Oh yes. Seems some financial users have strict admission policies for running as non-root or minimal permission mode. I think OpenShift also has similar requirements.

How about we run sn-platform in minimal permission mode by default, or add a flag to do so and generate these security context settings automaticallY/

charts/sn-platform/templates/vault/vault-statefulset.yaml
```diff
@@ -32,9 +32,22 @@ spec:
...
   serviceType: {{ .Values.vault.serviceType }}
   statsdDisabled: true
+  {{- if .Values.vault.securityContext }}
+  securityContext: {{- toYaml .Values.vault.securityContext | nindent 4 }}
+  vaultEnvsConfig:
+    - name: SKIP_CHOWN
+      value: "true"
+    - name: SKIP_SETCAP
+      value: "true"
+    - name: VAULT_DISABLE_MLOCK
```
maxsxu, line 42: Disabling mlock is not recommended: https://www.vaultproject.io/docs/configuration#disable_mlock

ericsyh: emmm, seems `mlock` needs root privilege. I got error message when I try to run vault as non-root

```
Error initializing core: Failed to lock memory: cannot allocate memory

This usually means that the mlock syscall is not available.
Vault uses mlock to prevent memory from being swapped to
disk. This requires root privileges as well as a machine
that supports mlock. Please enable mlock on your system or
disable Vault from using it. To disable Vault from using it,
set the `disable_mlock` configuration option in your configuration
file.
2022-08-11T01:26:29.465Z [INFO]  proxy environment: http_proxy="" https_proxy="" no_proxy=""
```

tuteng:
Please try rebase master for fixed test

tuteng:
@fantapsody @maxsxu @tuteng PTAL

charts/sn-platform/templates/vault/vault-statefulset.yaml
MarvinCai: specifying `securityContext` doesn't need to mean run as non-root user, so shall we just add a separate knob for these envs?

ericsyh:
Close PR for it is from fork.

PR #851, opened 2022-08-22
[Pulsar] Add oauth2 support for pulsar charts

### LINKED ISSUE #850: [Pulsar] Support oauth2 as the authentication provider
In some pulsar-related projects repo, developers may want to run ci against a pulsar with the oauth2 authentication provider, but there is no convenient way to deploy a Pulsar with oauth2 enabled

jiangpengcheng:
Fixes #850

### Motivation

Make it easy to test Pulsar with oauth2 enabled

### Modifications

Add the oauth2 authentication provider support to the Pulsar charts

### Verifying this change

### Documentation

Check the box below.

Need to update docs? 
  
- [x] `doc-required` 
  
  There is an example values file, but a doc would be better

### DIFF
--- a/examples/pulsar/values-oauth2.yaml
+++ b/examples/pulsar/values-oauth2.yaml
@@ -1,0 +1,47 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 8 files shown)


freeznet (approved):

examples/pulsar/values-oauth2.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+auth:
+  authorization:
+    # if set to true, please set one of the superRoles to your super-application-id
+    enabled: false 
+  authentication:
+    enabled: true
+    provider: "oauth2"
+    # below configurations are using Azure as the provider
+    # please refer to: https://streamnative.slab.com/posts/how-oauth-2-broker-plugin-work-with-azure-active-directory-azure-ad-platform-ui9a8ox8
```
maxsxu, line 37: I think we should remove these private links in this public repo

jiangpengcheng: ok, then I will add a doc in this pr to describe how to use azure as the oauth2 provider

maxsxu (approved):

PR #1111, opened 2023-10-13
Load the pulsar env as the program start options

zymap:
---

### Motivation

When enable tls for the zookeeper, it will write the TLS properties with PULSAR_EXTRA_OPTS as the key into the pulsar-env.sh file. The metadata tool also needs the PULSAR_EXTRA_OPTS to get the zookeeper TLS configurations. So import the PULSAR_EXTRA_OPTS into the OPTS to load the zookeeper TLS configuration.


zymap:
@ericsyh @maxsxu PTAL, thankx

ericsyh (approved):

PR #1175, opened 2024-06-18
Update jwt-secret-init-job.yaml

merlimat:
### Motivation

Use `/tmp/` directory for writing scripts, to avoid permission issues.

### Modifications

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/charts/sn-platform-slim/templates/toolset/jwt-secret-init-job.yaml
+++ b/charts/sn-platform-slim/templates/toolset/jwt-secret-init-job.yaml
@@ -64,18 +64,18 @@ spec:
         args:
           - |
             set -ex;
-            mkdir -p scripts/pulsar;
-            cp scripts/jwt-secret-config/* scripts/pulsar;
-            chmod +x scripts/pulsar/*;
+            mkdir -p /tmp/scripts/pulsar;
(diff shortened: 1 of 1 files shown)


nlu90 (approved):

lhotari:
> Use `/tmp/` directory for writing scripts, to avoid permission issues.

btw. Some environments might prevent script execution from /tmp with `noexec` mount option. That's rare, but I wouldn't be surprised if Fico has such settings. In those environments loading native libraries (for example tcnative) from /tmp is also prevented. I noticed that that [`readOnlyRootFilesystem` support in sn-operator also mounts /tmp to a custom emptyDir based volume](https://github.com/streamnative/sn-operator/blob/2ca388a8730e6b62831f32a5268bc2a810112690/pkg/commons/readonly/readonly.go#L55), perhaps to workaround the `noexec` issue. That's also one reason why it's worth using operator's `readOnlyRootFilesystem` setting in restricted environments.

PR #1282, opened 2026-09-15
feat: allow configuring ServiceAccount token automount

tuteng:

Fixes #<xyz>

Master Issue: #<xyz>

### Motivation

### Modifications

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

labuladong (dismissed):

fantapsody (commented):
Reviewed by rendering this branch with `helm template` against both `sn-platform` and `sn-platform-slim`. The mechanism looks right — the `kindIs "bool"` guard means the `null` default is a no-op, so existing deployments are unaffected, and setting the value to `false` renders `automountServiceAccountToken: false` on the correct object in every case I checked.

Two gaps before this fully closes the kyverno finding:

**1. Four ServiceAccounts aren't covered.** The policy scans every ServiceAccount in the namespace, so these will still be reported if the customer enables the components:

- `charts/sn-platform/templates/custom-metric-server/custom-metric-server.yaml`
- `charts/sn-platform/templates/custom-metric-server/custom-metric-server-init.yaml`
- `charts/sn-platform/templates/custom-metric-server/custom-metric-server-prometheus.yaml`
- `charts/{sn-platform,sn-platform-slim}/templates/control-center/ingress-controller-rbac.yaml` (gated on `ingress.controller.rbac`)

**2. No global switch.** As it stands the customer has to set 13+ separate per-component keys, and again for the other chart. A `global.serviceAccount.automountServiceAccountToken` that each component falls back to would turn their values override into one line. Something like:

```
{{- $automount := default .Values.global.serviceAccount.automountServiceAccountToken .Values.broker.serviceAccount.automountServiceAccountToken }}
{{- if kindIs "bool" $automount }}
automountServiceAccountToken: {{ $automount }}
{{- end }}
```

**Minor:** `zookeeper.customTools.{backup,restore}.serviceAccount.automountServiceAccountToken` is a new path, but the annotations on those same ServiceAccounts still come from `zookeeper.customTools.serviceAccount`. Two config paths for one object is easy to get wrong later — worth putting the automount setting alongside the annotations unless per-job control is deliberate.

One thing worth confirming with the customer before they roll this out: `external-dns`, `prometheus`, `vault` and `function-worker` all need API server access. `external-dns` on EKS is fine (IRSA injects its own projected token volume independently of `automountServiceAccountToken`), but `vault`'s Kubernetes auth and `prometheus`'s service discovery will break if automount is disabled for them without a replacement projected token.

labuladong (approved):

PR #1283, opened 2026-09-16
feat: chart workload security

tuteng:

Fixes #<xyz>

Master Issue: #<xyz>

### Motivation

### Modifications

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

labuladong (dismissed):

fantapsody (commented):
Reviewed by rendering this branch with `helm template` against both `sn-platform` and `sn-platform-slim`.

## Blocker: the slim detector template produces invalid YAML

`charts/sn-platform-slim/templates/detector/pulsar-detector-deployment.yaml` fails to render as soon as `pulsar_detector.extraVolumeMounts` **or** `pulsar_detector.extraVolumes` is set:

```
$ helm template t charts/sn-platform-slim --set pulsar_detector.extraVolumeMounts[0].name=tmp \
    --set pulsar_detector.extraVolumeMounts[0].mountPath=/extra
Error: YAML parse error on sn-platform-slim/templates/detector/pulsar-detector-deployment.yaml:
error converting YAML to JSON: yaml: line 129: did not find expected key
```

Two causes, both slim-only (`charts/sn-platform` renders fine with the same values):

- **L166–172** — `volumeMounts:` is now unconditional with its items at indent 8, but `extraVolumeMounts` is still emitted with `nindent 10`.
- **L180** — `extraVolumes` is still `nindent 8`. The equivalent line in `charts/sn-platform` was changed to `nindent 6` in this PR, but the fix wasn't mirrored here.

This sits directly on the new feature's happy path: the `adminTokenFile` doc comment tells users to mount the token through `extraVolumes` / `extraVolumeMounts`, which is exactly the combination that fails.

## The two charts have diverged

Whichever chart the customer is on, the fix should land in both. Current state:

| | `sn-platform-slim` | `sn-platform` |
|---|---|---|
| detector `readOnlyRootFilesystem` | **`true` by default** | absent |
| detector `/tmp` + `/pulsar/logs` mounts | added | not added |
| detector `extraVolumes` indent fix | ✗ | ✓ |
| jwt-init `kubectl` container | rewritten (`mkdir -p` + `/tmp` mount), sc from values | hardcoded sc, old `cp` + `/tmp/binaries` |
| `busybox` init container | `readOnlyRootFilesystem: true`, simplified copy | old `busybox --install`, no ro-rootfs |

The detector asymmetry is the risky one: slim flips `readOnlyRootFilesystem: true` **on by default** for every existing user of that chart, while the toolset deliberately stays opt-in (`readOnlyRootFilesystem: false` plus the new "enable explicitly after confirming all runtime write paths are mounted writable" comment). The same reasoning should apply to the detector — default it off, or at minimum apply the identical treatment (including the `/tmp` and `/pulsar/logs` mounts) to `sn-platform`, which today would break if anyone set it.

## Verified as fine

Noting these so they don't get re-litigated:

- `defaultMode: 0440` renders as `288` (`r--r-----`), and `streamnative_console.securityContext` carries `fsGroup: 1000`, so the non-root console process can still read the token files. It only breaks for someone who overrides `streamnative_console.securityContext` without `fsGroup` — probably worth a note in the values comment.
- The keytool `outputDir=/pulsar/conf` change for `toolset`: the only consumer of `/pulsar/toolset.truststore.jks` in the repo is `sn-platform/templates/toolset/toolset-configmap.yaml`, which is updated. No other references, and the other components keep `/pulsar`. The `kafka` sidecar writes into its own container's `/pulsar/conf` and doesn't get `readOnlyRootFilesystem`, so that path still works.
- `mergeOverwrite (deepCopy .Values.toolset.containerSecurityContext) (dict "readOnlyRootFilesystem" ...)` behaves correctly, and the new `allowPrivilegeEscalation` / `capabilities.drop` values land on `containerSecurityContext`, not the pod-level block.
- The `adminTokenFile` shell quoting renders correctly and the `secretKeyRef` env injection is properly suppressed.

## One question

Neither PR adds `seccompProfile: RuntimeDefault`, which is normally part of the same restricted-PSS kyverno bundle. If FICO's policy set includes it, that's the next ticket — worth confirming now rather than after another round.

labuladong (approved):

